PR #7, opened 2016-12-15
remove PENDING_CLOSE flag

wei-lee:
We have come across a problem when try to use this module. We found that the circuit breaker will be kept open if it's going to the "halfOpen" state the second time. There will be no way to close the circuit breaker afterwards.

After a little of debugging, it looks like the problem is caused by the `PENDING_CLOSE` flag. I can't really tell why we need the flag, and removing it fix the problem. 

The case is being shown in the test.

@lance can you please review this and see if I am right? 

Thanks.

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -7,7 +7,6 @@ const STATE = Symbol('state');
 const OPEN = Symbol('open');
 const CLOSED = Symbol('closed');
 const HALF_OPEN = Symbol('half-open');
-const PENDING_CLOSE = Symbol('pending-close');
 const FALLBACK_FUNCTION = Symbol('fallback');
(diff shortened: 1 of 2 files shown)


lance:
@wei-lee thanks for making the contribution. But we actually do need `PENDING_CLOSE`.

It's a real edge case, but the reason we have `PENDING_CLOSE` is because the circuit should only ever attempt to execute once when in the `HALF_OPEN` state. In a scenario where lots and lots of requests are occurring in parallel, many could execute while the circuit is `HALF_OPEN`. By checking the `PENDING_CLOSE` state, we ensure this doesn't happen.

You _did_ discover a bug, however. So thank you! The `PENDING_CLOSE` flag was never being reset. I've fixed that, and used your test case in my latest commit.

PR #9, opened 2016-12-21
build: Ignoring generated file browser-test.js

helio-frota:


helio-frota:
Closing since @lance will change the strategy related to this file.

PR #10, opened 2016-12-22
bug: removing unused util import in jquery example

lholmquist:
i'm not sure if bug is the correct label here


lance:
Thanks @lholmquist

PR #22, opened 2017-03-03
enhancement: prepend the circuit events to see the most recent event …

lholmquist:
…first

This is a small change to how the circuit events are listed in the UI.  Showing the most recent event on the top makes it a little easier to see what is going on without having to scroll to the bottom of the page. then scroll back up again to click the button.


lance:
Thanks @lholmquist!

PR #30, opened 2017-03-07
feat: Add caching capability to circuits

helio-frota:


lib/status.js
```diff
@@ -33,13 +33,23 @@ class Status {
      * The number of times this circuit breaker has timed out
      */
     this.timeouts = 0;
+    /**
+     * The number of times the cache is used
+     */
+    this.cacheHit = 0;
```
lance: Even though I'm not very fond of the naming scheme I originally chose, I think we should be consistent and make this and `cacheMiss` plural: `cacheHits` and `cacheMisses`.

lib/circuit.js
```diff
@@ -222,6 +232,17 @@ class CircuitBreaker extends EventEmitter {
         handleError(error, this, timeout, args, resolve, reject);
       }
     });
+    if (this.options.cache) {
```
lance: Is it overkill to allow the option to be either a `boolean`, an `Object`, a `Map` or a `WeakMap`? This would allow users to create a circuit with a pre-primed cache. 

As I'm writing this, I am leaning towards that being a feature, but perhaps a feature for later... :)

lib/circuit.js
```diff
@@ -222,6 +232,17 @@ class CircuitBreaker extends EventEmitter {
...
+      CACHE.set(this, value);
```
lance: Hmm. This would mean that we are caching failures as well, I suppose. I'm not sure I like that.

lib/circuit.js
```diff
@@ -165,6 +169,12 @@ class CircuitBreaker extends EventEmitter {
    * @fires CircuitBreaker#timeout
    */
   fire () {
+    if (CACHE.get(this) !== undefined) {
+      this.emit('cacheHit');
+      return CACHE.get(this);
+    } else {
+      this.emit('cacheMiss');
```
lance: Hmm. The `cacheMiss` event would fire here even if caching were not enabled.

lance: Hmm. I say "hmm" a lot.

lib/circuit.js
helio-frota: true, caching the `returned value`. failure or not.

lib/circuit.js
helio-frota: true! I'll fix this.

lib/circuit.js
helio-frota: I'm not following : ]
if the option.cache === Object then this will be true on this if statement...

lance: Yeah, my comment is in the wrong line of code. Essentially, I was saying, in the ctor, we could check the type of object. If boolean, use our WeakMap - if Object, Map or WeakMap, use what's provided. In either case, set a flag that the check here would use. Make sense?

helio-frota: perhaps a feature for later ? :D

test/test.js
```diff
@@ -40,6 +59,35 @@ test('Passes parameters to the circuit function', (t) => {
...
+  };
+  const breaker = cb(passFail, options);
+
+  breaker.fire(expected)
+    .then((arg) => t.equals(arg, expected, `cache hits:misses ${breaker.status.cacheHits}:${breaker.status.cacheMisses}`))
+    .catch(t.fail)
+    .then(() => {
+      breaker.fire(expected)
+        .then((arg) => {
+          t.equals(arg, expected, `cache hits:misses ${breaker.status.cacheHits}:${breaker.status.cacheMisses}`);
```
lance: You should add a test assertion here that ensures there are 0 cache hits and 1 cache miss.

test/test.js
```diff
@@ -40,6 +59,35 @@ test('Passes parameters to the circuit function', (t) => {
...
+          breaker.clearCache();
+        })
+        .catch(t.fail)
+        .then(() => {
+          breaker.fire(expected)
+            .then((arg) => {
+              t.equals(arg, expected, `cache hits:misses ${breaker.status.cacheHits}:${breaker.status.cacheMisses}`);
```
lance: Similar to above, there should be a check here for 1 cache hit and 1 cache miss.

PR #34, opened 2017-03-08
feat: circuit status now contains a rolling window

lance:
## This is an ongoing PR, do not merge yet

The rolling stats window is configurable in both total time sampled, and
how many snapshots, or buckets, within that time frame.

The new options that can be applied to a circuit breaker are
`rollingCountTimeout` and `rollingCountBuckets`, which default to 10000
and 10, respectively. So by default, the window is a statistical view
over the last 10 seconds, consisting of 10 one second snapshots.

The current circuit breaker status api has not been modified directly,
however the expected results are different. For example, you can still
do this:

```js
console.log(`Failure count: ${circuit.status.failures}`);
```

But that count will consist only of the number of failures within the
current snapshot.

To obtain stats for the entire window, use the `window` property.

```js
const stats = circuit.status.window;
```

This will give you an array containing the statistical sampling for the
entire window. So, given the defaults noted above, by default this will
be a ten element array, with each element containing an object with
sample data that looks something like this:

```js
{
  failures: 11,
  fallbacks: 9,
  successes: 3491,
  rejects: 2,
  fires: 3493,
  timeouts: 0,
  start: 1488999002013
}
```

### DIFF
--- a/lib/status.js
+++ b/lib/status.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const CIRCUIT_BREAKER = Symbol('circuit-breaker');
+const STATS_WINDOW = Symbol('stats-window');
 
 /**
  * @class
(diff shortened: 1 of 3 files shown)


lib/status.js
```diff
@@ -9,45 +10,101 @@ const CIRCUIT_BREAKER = Symbol('circuit-breaker');
...
    * The number of times this circuit breaker has timed out
+   * during the current statistical window.
    */
-  status.timeouts = 0;
+  get timeouts () {
+    return this[STATS_WINDOW][0].timeouts;
+  }
 }
 
+const stats = () => ({
```
lance: I will merge from master before pushing my changes. There will be conflicts in this file, but I will deal with them.

lance:
Merging this as is for the time being and adding #37 #38 and #39 as new issues.

Issue #39, opened 2017-03-30
A circuit's status data should be provided as a data stream


lance:
Specifically, implement https://nodejs.org/api/stream.html#stream_class_stream_readable

lholmquist:
on this now.  looking to https://github.com/awolden/brakes/blob/master/lib/globalStats.js for inspiration

lholmquist:
added here  2d44df68f549b6b8149f49322f8b9cc7e32cc9c9

PR #41, opened 2017-03-31
Addition of streaming hystrix data.

lholmquist:
DO NOT MERGE - this is a WIP for #39 

I've added a Class, `HystrixStats`, which produces a `stream` of the hystrix metrics.

A circuit breaker will `register` with this class during the creation of a circuit breaker, https://github.com/lholmquist/opossum/blob/status-stream-data/lib/circuit.js#L101 so for every snapshot event emitted, the stats stream is updated.

I'm storing the instance of the circuit that is passed in, in a WeakMap so the hystrix event stream can get a handle on it to give me the circuits name, group, etc..

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,11 @@ you create the breaker. E.g.
 // Force opossum to use native JS promises
 const breaker = circuitBreaker(readFile, { Promise: Promise });
 ```
+
+### Hystrix Metrics
+
+A Hystrix Stream is available for use with a Hystrix Dashboard using the `circuitBreaker.hystrixStats.getHystrixStream` method.
+
+This method returns a [Node.js Stream](https://nodejs.org/api/stream.html), which makes it easy to create an SSE stream that will be compliant with a Hystrix Dashboard.
+
+Additional Reading: [Hystrix Metrics Event Stream](https://github.com/Netflix/Hystrix/tree/master/hystrix-contrib/hystrix-metrics-event-stream), [Turbine](https://github.com/Netflix/Turbine/wiki), [Hystrix Dashboard](https://github.com/Netflix/Hystrix/wiki/Dashboard)
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -2,6 +2,7 @@
 
 const EventEmitter = require('events');
 const Status = require('./status');
+const HystrixStats = require('./hystrix-stats');
 
 const STATE = Symbol('state');
 const OPEN = Symbol('open');
@@ -12,6 +13,7 @@ const FALLBACK_FUNCTION = Symbol('fallback');
 const STATUS = Symbol('status');
 const NAME = Symbol('name');
 const GROUP = Symbol('group');
+const HYSTRIX_STATS = Symbol('hystrix-stats');
 const CACHE = new WeakMap();
 
 /**
@@ -95,6 +97,9 @@ class CircuitBreaker extends EventEmitter {
     if (this.options.cache) {
       CACHE.set(this, undefined);
     }
+
+    // Register this instance of the circuit breaker with the hystrix stats listener
+    this[HYSTRIX_STATS] = new HystrixStats(this);
   }
 
   /**
@@ -179,6 +184,12 @@ class CircuitBreaker extends EventEmitter {
     return this[STATUS].stats;
   }
 
+  /**
+    A convenience function that returns the hystrixStats
+  */
+  get hystrixStats () {
+    return this[HYSTRIX_STATS];
+  }
   /**
    * Provide a fallback function for this {@link CircuitBreaker}. This
    * function will be executed when the circuit is `fire`d and fails.
--- a/lib/hystrix-formatter.js
+++ b/lib/hystrix-formatter.js
@@ -1,0 +1,69 @@
+'use strict';
+
+// A function to map our stats data to the hystrix format
+// returns JSON
+function hystrixFormatter (stats) {
+  const json = {};
+  json.type = 'HystrixCommand';
+  json.name = stats.name;
+  json.group = stats.group;
+  json.currentTime = new Date();
+  json.isCircuitBreakerOpen = !stats.closed;
+  json.errorPercentage = stats.fires === 0 ? 0 : (stats.failures / stats.fires) * 100;
+  json.errorCount = stats.failures;
+  json.requestCount = stats.fires;
+  json.rollingCountBadRequests = stats.failures;
+  json.rollingCountCollapsedRequests = 0;
+  json.rollingCountEmit = stats.fires;
+  json.rollingCountExceptionsThrown = 0;
+  json.rollingCountFailure = stats.failures;
+  json.rollingCountFallbackEmit = stats.fallbacks;
+  json.rollingCountFallbackFailure = 0;
+  json.rollingCountFallbackMissing = 0;
+  json.rollingCountFallbackRejection = 0;
+  json.rollingCountFallbackSuccess = 0;
+  json.rollingCountResponsesFromCache = stats.cacheHits;
+  json.rollingCountSemaphoreRejected = stats.rejects;
+  json.rollingCountShortCircuited = stats.rejects;
+  json.rollingCountSuccess = stats.successes;
+  json.rollingCountThreadPoolRejected = 0;
+  json.rollingCountTimeout = stats.timeouts;
+  json.currentConcurrentExecutionCount = 0;
+  json.rollingMaxConcurrentExecutionCount = 0;
+  // TODO: caluclate these latency values
+  json.latencyExecute_mean = 0;
+  json.latencyExecute = {
+    '0': 0,
+    '25': 0,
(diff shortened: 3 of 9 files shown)


lib/circuit.js
```diff
@@ -93,6 +94,9 @@ class CircuitBreaker extends EventEmitter {
     if (this.options.cache) {
       CACHE.set(this, undefined);
     }
+
+    // Register this instance of the circuit breaker with the hystrix stats listener
+    hystrixStats.register(this[STATUS]);
```
lance: With my [latest push](https://github.com/bucharest-gold/opossum/commit/7bb3931c9df3df80cde3cb54cdae5cf6ef18b4e0), this should be `this[STATUS].stats`. That will return an `Object` that has the properties you can see [here](https://github.com/bucharest-gold/opossum/blob/master/lib/status.js#L117-L126). It represents the cumulative data for each event over the entire window.

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+// A function to map our stats data to the hystrix format
+// returns JSON
+function hystrixFormatter (stats) {
+  const json = {};
+  json.type = 'HystrixCommand';
+  json.name = 'ola';
```
lance: The circuit breaker now has a name property that you can use for this.

lib/hystrix-stats.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  // The stats coming in should be already "Reduced" and in
+  _hystrixTransformer (stats, encoding, cb) {
+    console.log('transformer', stats, encoding);
+
+    const formattedStats = hystrixFormatter(stats);
+
+    // Need to take the stats and map them to the hystrix format
+    return cb(null, `data: ${JSON.stringify(formattedStats)}\n\n`);
```
lance: Isn't `hystrixFormatter` already returning JSON?

lib/hystrix-stats.js
```diff
@@ -0,0 +1,61 @@
...
+  // This will take the stats data from the listener and push it on the stream to be transformed
+  _hystrixSnapshotListener (stats) {
+    console.log('_hystrixSnapshotListener', stats);
+    this._readableStream.push(stats);
+  }
+
+  // Register the instance of a circuit to listen for a snapshot event
+  // during that event, call the hystrixSnapshotListener to do the stream stuff
+  register (circuitInstance) {
+    circuitInstance.on('snapshot', this._hystrixSnapshotListener.bind(this));
```
lance: Currently it's the `Status` object that emits `snapshot`. That could change if you think it's better suited in `CircuitBreaker`, but I think it would be unnecessary indirection. It would mean that the circuit breaker listens to the status object for snapshot events and just propagates them to the client.

lib/hystrix-stats.js
lholmquist: it does,  this is turning it into a string for the event stream

lib/hystrix-formatter.js
lholmquist: I'm not sure if i can get access to this information currently.  Since the Status object emits the snapshot event,  i only get the stats data during that event.   maybe the Circuit class should emit the event(which i'm not sure about), or we add some metadata to the stats object that gets emitted during that snapshot event

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
...
+    '95': 0,
+    '99': 0,
+    '99.5': 0,
+    '100': 0
+  };
+  json.latencyTotal_mean = 0;
+  json.latencyTotal = { '0': 0, '25': 0, '50': 0, '75': 0, '90': 0, '95': 0, '99': 0, '99.5': 0, '100': 0 };
+  json.propertyValue_circuitBreakerRequestVolumeThreshold = 5;
+  json.propertyValue_circuitBreakerSleepWindowInMilliseconds = 5000;
+  json.propertyValue_circuitBreakerErrorThresholdPercentage = 50;
```
lance: You should use `circuit.options` to get this value as well as the previous two

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
...
+  json.propertyValue_circuitBreakerForceOpen = false;
```
lance: We have functions to force open/closed a circuit. Maybe we should add an issue for the circuit to keep track of this so it can be reported.

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
...
+  json.propertyValue_circuitBreakerForceClosed = false;
+  json.propertyValue_circuitBreakerEnabled = true;
```
lance: Use `circuit.closed` here, maybe? I'm not sure if this stat is supposed to reflect the current state of the circuit breaker, or if with Hystrix, a user could enable/disable a circuit.

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
...
+  json.propertyValue_executionIsolationStrategy = 'THREAD';
+  json.propertyValue_executionIsolationThreadTimeoutInMilliseconds = 300;
+  json.propertyValue_executionTimeoutInMilliseconds = 300;
```
lance: Use the circuit's options here as well.

lib/hystrix-formatter.js
```diff
@@ -0,0 +1,68 @@
...
+  json.propertyValue_executionIsolationThreadInterruptOnTimeout = true;
+  json.propertyValue_executionIsolationThreadPoolKeyOverride = null;
+  json.propertyValue_executionIsolationSemaphoreMaxConcurrentRequests = 10;
+  json.propertyValue_fallbackIsolationSemaphoreMaxConcurrentRequests = 10;
+  json.propertyValue_metricsRollingStatisticalWindowInMilliseconds = 10000;
+  json.propertyValue_requestCacheEnabled = true;
```
lance: And here :)

lib/hystrix-stats.js
```diff
@@ -0,0 +1,62 @@
...
+
+  // This will take the stats data from the listener and push it on the stream to be transformed
+  _hystrixSnapshotListener (stats) {
+    const circuit = this.circuitInstanceMap.get(this);
+    this._readableStream.push(Object.assign({}, {name: circuit.name, closed: circuit.closed, group: circuit.group}, stats));
+  }
+
+  // Register the instance of a circuit stats to listen for a snapshot event
+  // during that event, call the hystrixSnapshotListener to do the stream stuff
+  register (circuitInstance) {
```
lance: Is there any reason to have this separated out? Couldn't the user just pass the circuit to the constructor, and this code could be executed there, no?

lholmquist: currently the class is exported as a new instance,  so you wouldn't be able to pass it along since the constructor is called while you require it.  

I think i do like that better though,  having the CircuitBreakers constructor calling `new HystrixStats()` and registering then,   let me change that

lib/hystrix-formatter.js
lholmquist: that sounds like a good idea

lib/hystrix-formatter.js
lholmquist: that last property i think i found in the options,  but what would those other 2 be,  the rollingCountTimeout and buckets?

lib/hystrix-formatter.js
lholmquist: looking at what the brakes library is doing for this value, they have it hard coded to true and have a comment about it being "not reported"

I can check the java lib to see what they do with it

lib/hystrix-formatter.js
lholmquist: i've created an issue https://github.com/bucharest-gold/opossum/issues/46

lib/hystrix-formatter.js
lholmquist: looking at the java lib, its here:  https://github.com/Netflix/Hystrix/blob/master/hystrix-core/src/main/java/com/netflix/hystrix/HystrixCommandProperties.java#L69

The comment is "// Whether circuit breaker should be enabled.", so this makes me think it doesn't have to do with the circuits closed/open state, but if the circuit breaker should be used or not

lib/hystrix-formatter.js
lance: @lholmquist `circuitBreakerSleepWindowInMilliseconds` should be our `options.resetTimeout`. I guess we don't measure request volume, so we don't have a property for it. :)

lib/hystrix-formatter.js
lance: Makes sense

README.md
```diff
@@ -231,3 +231,11 @@ you create the breaker. E.g.
 // Force opossum to use native JS promises
 const breaker = circuitBreaker(readFile, { Promise: Promise });
```
lance: I'm debating removing Fidelity from opossum, fwiw.

Issue #43, opened 2017-04-03
Circuit needs a Group property


lholmquist:
One of the properties in the Hystrix Data is "Group",  https://github.com/Netflix/Hystrix/tree/master/hystrix-contrib/hystrix-metrics-event-stream

We can probably default this to be the same as the name property if nothing is passed in

lance:
This can be closed, can't it?

lholmquist:
Yup,  just looking for the commit.

closed via 3052f23af43c98a622bf51b25edcef969d9b39c0

PR #44, opened 2017-04-03
feat: add a group option.  GH-Issue #43

lholmquist:
adds the group option and defaults to use the name property if none is provided


lholmquist:
@lance thoughts?

PR #55, opened 2017-04-11
WIP - first pass at percentiles

lholmquist:
connects to https://github.com/bucharest-gold/opossum/issues/38

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -39,6 +39,9 @@ const CACHE = new WeakMap();
  *  10000, and options.rollingCountBuckets is 10, then the statistical window
  *  will be 1000 1 second snapshots in the statistical window. Default: 10
  * @param options.name the name to use for this circuit when reporting stats
+ * @param options.rollingPercentilesEnabled {boolean} This property indicates whether execution latencies
+ *  should be tracked and calculated as percentiles.
+ *  If they are disabled, all summary statistics (mean, percentiles) are returned as -1.
  */
 class CircuitBreaker extends EventEmitter {
   constructor (action, options) {
@@ -46,6 +49,7 @@ class CircuitBreaker extends EventEmitter {
     this.options = options;
     this.options.rollingCountTimeout = options.rollingCountTimeout || 10000;
     this.options.rollingCountBuckets = options.rollingCountBuckets || 10;
+    this.options.rollingPercentilesEnabled = options.rollingPercentilesEnabled !== false;
 
     this[STATUS] = new Status(this.options);
     this[STATE] = CLOSED;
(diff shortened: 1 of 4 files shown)


lholmquist:
@lance @helio-frota mind just taking a quick look at this if you guys have a chance, just to see if the code looks legit

lib/status.js
```diff
@@ -81,12 +86,45 @@ class Status extends EventEmitter {
...
       if (!val) return acc;
-      Object.keys(acc).forEach(key => (acc[key] += val[key] || 0));
+      Object.keys(acc).forEach(key => {
+        if (key !== 'latencyTimes' && key !== 'percentiles') {
+          (acc[key] += val[key] || 0);
+        }
+      });
+
+      if (this.rollingPercentilesEnabled) {
+        acc.latencyTimes.push.apply(acc.latencyTimes, val.latencyTimes || []);
```
lance: Why not just call `push()`? What's the need for `apply`?

lholmquist: we need to concatenate the values in that `latencyTimes` so if we just did push, then we would get something like this: `[ [1], [2], [3], [4] ]` instead of `[1,2,3,4]` .   would be nice if we could use the spread operator here, but i don't think it goes back to node 4

lib/circuit.js
```diff
@@ -275,8 +282,9 @@ class CircuitBreaker extends EventEmitter {
            * Emitted when the circuit breaker action takes longer than `options.timeout`
            * @event CircuitBreaker#timeout
            */
-          this.emit('timeout', error);
-          resolve(fallback(this, error, args) || fail(this, error, args));
+          const latencyEndTime = Date.now() - latencyStartTime;
```
lholmquist: yes,  that makes sense

lib/circuit.js
```diff
@@ -316,15 +327,15 @@ class CircuitBreaker extends EventEmitter {
...
-  fail(circuit, error, args);
-  const fb = fallback(circuit, error, args);
+  fail(circuit, error, args, latencyEndTime);
+  const fb = fallback(circuit, error, args, latencyEndTime);
   if (fb) resolve(fb);
   else reject(error);
 }
 
-function fallback (circuit, err, args) {
+function fallback (circuit, err, args, latencyEndTime) {
```
lholmquist: ah,  doesn't look like it is,  i think i put it in there because i had a thought about storing the latency during the fallback event, but i don't think we need to do that,  i'll remove it for now

lance:
Hmm - I think ultimately we probably want to measure fallback latency as
well. Maybe we can add that as another issue.

test/test.js
```diff
@@ -647,6 +647,61 @@ test('options.maxFailures should be deprecated', (t) => {
...
+  const breaker = cb(passFail);
+  breaker.on('success', (result, latencyTime) => {
+    t.ok(latencyTime, 'second argument is the latency');
+    t.end();
+  });
+
+  breaker.fire(1);
+});
+
+test('Circuit Breaker timeout event emits latency', (t) => {
```
lance: This should say "Breaker failure event". And the one below should say "Breaker timeout event" :-)

Issue #57, opened 2017-04-20
Stuck in an open circuit breaker state


lholmquist:
While i was recording the screencast showing how to use opossum, i sort of got myself stuck in an always opened state.  

I'll try to reproduce this with a test,  if not i can do a screencast of what i encountered.

PR #59, opened 2017-04-20
fix: don't let circuits get stuck half open

lance:
Fixes: https://github.com/bucharest-gold/opossum/issues/57
If a circuit is fired while half open AND while PENDING_CLOSE,
it gets stuck there. This fixes that.

This work connects to #57


lance:
@lholmquist can you see if this fixes your issue?

lholmquist:
landed 4d89ae4

PR #71, opened 2017-06-13
Added semaphore and blue-test for specific failed test.

helio-frota:


lib/circuit.js
```diff
@@ -235,89 +238,92 @@ class CircuitBreaker extends EventEmitter {
...
-      return CACHE.get(this);
-    } else if (this.options.cache) {
-      /**
-       * Emitted when the circuit breaker does not find a value in
-       * the cache, but the cache option is enabled.
-       * @event CircuitBreaker#cacheMiss
-       */
-      this.emit('cacheMiss');
-    }
+    const semaphore = new Semaphore(this.options.capacity);
```
lance: This should be a class property. Creating a new semaphore each time the circuit fires doesn't achieve the goal of rate limiting, because the semaphore is only ever used once, and cannot block other code paths from acquiring it.

lib/circuit.js
```diff
@@ -235,89 +238,92 @@ class CircuitBreaker extends EventEmitter {
...
-          if (!timeoutError) {
-            clearTimeout(timeout);
+        return fallback(this, 'Breaker is open', args, 0) ||
+          Promise.reject(new Error('Breaker is open'));
+      }
+      this[PENDING_CLOSE] = false;
+
+      let timeout;
+      let timeoutError = false;
+      return new Promise((resolve, reject) => {
```
lance: I think we don't need to call semaphore.use() until here. All of the other stuff should be allowed to execute. E.g. I think it's safe to check for a cache hit and also to check if the circuit is open before attempting to acquire the semaphore. No reason for that stuff to be rate limited.

helio-frota:
```
helio-frota> I'll close the PR on opossum and see what happened, I saw that the comments are changed, and was not me
<helio-frota> I think I forgot to update before send
<helio-frota> Emitted when the circuit breaker is open and failing fast
<helio-frota> Emitted when the circuit breaker action is executed
<helio-frota> going to close the PR and send again this is not part of my intention
```

helio-frota:
> real change is simply wrapping all of the code from fire() within a semaphore.use() scope. Is that the case?

Yes.

lib/circuit.js
helio-frota: Ok! Will change this. Thanks !

helio-frota:
Changes and new PR here https://github.com/bucharest-gold/opossum/pull/72

Issue #60, opened 2017-04-30
Use semaphore isolation to limit concurrent requests


lance:
## Fine Grained Bulkhead Pattern

The goal here is to implement a fine-grained bulkhead pattern similar to how Hystrix works in Java. This works hand-in-hand with circuit breakers to provide user-configurable client-side load balancing / throttling and helps to mitigate DoS attacks.

Before actually executing any of the circuit breaker wrapped code, the `fire()` function should acquire a semaphore, execute the function, and then release the semaphore. This prevents a client from unintentionally overloading a server with requests to a single resource, protecting against both programmer error (e.g. a loop that just spins, firing request after request), or malicious code designed to bring a service down.

### Things to Consider
* We need to find a reasonable default number of concurrent connections.
* We need to decide if we should check the semaphore first, or if we should check the open state first. If the circuit is open, we are essentially throttling anyway, so we don't need to acquire a semaphore, and we're not using any additional external resources because we are failing fast. 
* Does this limit also affect the fallback functions? If we decide that we check open state first, and potentially fallback, should the fallback functions be subject to the same rate limiting behavior.

lance:
https://www.npmjs.com/package/semaphore

helio-frota:
@lance I'm using https://www.npmjs.com/package/await-semaphore around the fire function.
Well, at least now I can run the tests and also this automatically runs the "leave".
But, I'm trying to understand to solve one issue with one of the tests (see on the screenshots ):

![a](https://user-images.githubusercontent.com/6443576/27046983-d4ff8eec-4f7c-11e7-9283-5791fb089d75.png)
![b](https://user-images.githubusercontent.com/6443576/27046984-d50057fa-4f7c-11e7-80a5-ecdbb978b104.png)

helio-frota:
I'm going to try again with semaphore.. this other library has some weird behavior I dunno.
When I change the capacity nothing happens, 0, 1, 100...

lance:
OK - if you are still having issues, send a link to your branch and I will
poke around.

helio-frota:
@lance https://github.com/bucharest-gold/opossum/tree/opossum-60

PR #72, opened 2017-06-13
feat: semaphore added

helio-frota:
Also using blue-tape to fix a specific test and
Connects to #60

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 node_js:
   - "4"
   - "6"
-  - "8"
+  - "8.1.0"
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 node_js:
   - "4"
   - "6"
-  - "8.1.0"
+  - "8"
 before_script:
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Here are the events you can listen for.
 * `close` - emitted when the breaker state changes to `closed`
 * `halfOpen` - emitted when the breaker state changes to `halfOpen`
 * `fallback` - emitted when the breaker has a fallback function and executes it
+* `semaphore-locked` - emitted when the breaker is at capacity and cannot execute the request
 
 Handling events gives a greater level of control over your application behavior.
 
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -3,6 +3,7 @@
 const EventEmitter = require('events');
 const Status = require('./status');
 const HystrixStats = require('./hystrix-stats');
+const Semaphore = require('await-semaphore').Semaphore;
 
 const STATE = Symbol('state');
 const OPEN = Symbol('open');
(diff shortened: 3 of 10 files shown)


helio-frota:
> Does it make sense to always use execution isolation with semaphores? What if there's not really a need for it? Why have it be always on? Reading the Netflix documentation here https://github.com/Netflix/Hystrix/wiki/How-it-Works#semaphores it really only makes sense to use semaphore isolation in certain use cases where the latency is very low and throughput requirements very high. Given that, we have no real way of implementing thread pool isolation/rate limiting, so maybe it does make sense to have it always on. I'm not 100% sure, but I think we should discuss it.

Indeed, also I need to research more about it.

> What happens to the semaphore when a circuit fires, but then times out? Is it released? We need to know that it is.

Good question, I'm going to investigate it.

> What happens when the semaphore has reached its limit and semaphore.acquire() is called? Does it return a rejected promise? Immediately? Or something else? 

> We should test for this.

Agree.

> The docs should be updated.

yup

helio-frota:
> What happens to the semaphore when a circuit fires, but then times out? Is it released? We need to know that it is.

seems like It is not releasing

```
 Circuit Breaker timeout with semaphore
    ✖ should be equal
    ------------------
      operator: equal
      expected: 10
      actual:   9
```

helio-frota:
@lanceball: not release when CB timeout is a blocker right ?
if yes (I think it is) then maybe we need to test another semaphore lib, because to call the release() we need first the the aquire().then(release => release()) ...

Also, I'm going try to change sem.use() to sem.aquire() and try to call the release() in some place around that timeout code

helio-frota:
I'm doing another test, but I missed to add more `released()` calls.
I'll push the changes soon.

helio-frota:
> What happens when the semaphore has reached its limit and semaphore.acquire() is called? Does it return a rejected promise? Immediately? Or something else?

> We should test for this.

So after change the code , and adding the `release()` calls, I noticed that the counter is always zeroed after release () and subsequent fire calls are not affected.

I will send the test code because I may be testing the wrong way. So it's good to have a new view on this part

helio-frota:
Now I'm using this lib to run the code in parallel :
https://www.npmjs.com/package/run-parallel

interesting because the code "runs" without error.

helio-frota:
@lance when running in parallel we have this :

```
  Semaphore capacity limit in parallel

    ✔ semaphore count is: 0 and initial capacity is: 1
    ✔ failures: 0
    ✔ fallbacks: 0
    ✔ successes: 2
    ✔ rejects: 0
    ✔ fires: 4
    ✔ semaphore count is: 0 and initial capacity is: 1
    ✔ failures: 0
    ✔ fallbacks: 0
    ✔ successes: 3
    ✔ rejects: 0
    ✔ fires: 4
    ✔ semaphore count is: 1 and initial capacity is: 1
    ✔ failures: 0
    ✔ fallbacks: 0
    ✔ successes: 4
    ✔ rejects: 0
    ✔ fires: 4
    ✔ This is not in parallel -> semaphore count is: 1 and initial capacity is: 1
    ✔ failures: 0
    ✔ fallbacks: 0
    ✔ successes: 4
    ✔ rejects: 0
    ✔ fires: 4
```

helio-frota:
> Given that, we have no real way of implementing thread pool isolation/rate limiting, so maybe it does make sense to have it always on. I'm not 100% sure, but I think we should discuss it.

true, with the current code we can see some real changes only when we run the code in parallel.
otherwise, looks like a serial code - like running without semaphore. Since it will release() every time btw.

PR #76, opened 2017-06-21
Add health check function to a circuit breaker

lance:
## Add `circuit.healthCheck(func[, interval])`

Exposes a way for users to provide their own health check function. The provided function
should return a promise. If the promise is rejected, the circuit will open and the `health-check-failed` event will be emitted.

## Usage
```js
const circuit = opossum(myFunction);
circuit.healthCheck(_ => {
    // check the status of the remote service
    if (remoteServiceDown) {
       return Promise.reject('Cannot connect to remote service');
    }
    return Promise.resolve();
  }, 10000); // check interval - default is 5000

circuit.on('health-check-failed', (err) => {
  // send a text message to the poor sysadmin
  sendErrorText(err);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Here are the events you can listen for.
 * `halfOpen` - emitted when the breaker state changes to `halfOpen`
 * `fallback` - emitted when the breaker has a fallback function and executes it
 * `semaphore-locked` - emitted when the breaker is at capacity and cannot execute the request
+* `health-check-failed` - emitted when a user-supplied health check function returns a rejected promise
 
 Handling events gives a greater level of control over your application behavior.
 
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -348,6 +348,49 @@ class CircuitBreaker extends EventEmitter {
   clearCache () {
     CACHE.set(this, undefined);
   }
+
+  /**
+   * Provide a health check function to be called periodically. The function should
+   * return a Promise. If the promise is rejected, the circuit will open. This is
+   * in addition to the existing circuit behavior as defined by `options.errorThresholdPercentage`
+   * in the constructor. For example, if the health check function provided here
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -151,6 +151,7 @@ Here are the events you can listen for.
 * `halfOpen` - emitted when the breaker state changes to `halfOpen`
 * `fallback` - emitted when the breaker has a fallback function and executes it
 * `semaphore-locked` - emitted when the breaker is at capacity and cannot execute the request
+* `health-check-failed` - emitted when a user-supplied health check function returns a rejected promise
```
lholmquist: I wonder if it also makes sense to have a Health check pass event.

I have no argument for or against :)

lib/circuit.js
```diff
@@ -348,6 +348,50 @@ class CircuitBreaker extends EventEmitter {
...
+   * @fires CircuitBreaker#health-check-failed
+   * @throws TypeError if `interval` is supplied but not a number
+   */
+  healthCheck (func, interval) {
+    interval = interval || 5000;
+    if (typeof func !== 'function') throw new TypeError('Health check function must be a function');
+    if (isNaN(interval)) throw new TypeError('Health check interval must be a number');
+
+    const check = _ => {
+      func.apply(this).catch(e => {
```
lholmquist: related to my comment about the health-check-passed event, should we do something on non error

README.md
lance: I thought about it, but it seems like a lot of unnecessary noise. Since there are only two possible states, pass or fail, it seems like the absence of fail implies pass.

lib/circuit.js
lance: Kind of have the same reasoning about this as I do about `health-check-passed`.

README.md
lholmquist: Yea, probably no reason to have it.  I'd say merge away 🎉

PR #80, opened 2017-07-05
Add failure check function to a circuit breaker

Permagate:
## Add circuit.failureCheck(func)

Exposes a way to separate which errors should be considered a failure that counts toward tripping the breaker. If the function returns true, then it should count as a failure. I find that I need this feature since I'd like to wrap some http functions with circuit breaker, but only want to trip on certain http status codes.

## Usage

```javascript
const circuit = opossum(myHttpFunction)

circuit.failureCheck(e => {
  // Only consider error with status code >= 500 as failures
  return e.statusCode >= 500;
})

circuit.fire();
```

## To-Do

* [ ] Update README

## Things to Consider

* Currently if the function returns false, I simply emit success with the error as an argument. Anything else that should be done?

Thanks beforehand 🙇

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -11,6 +11,7 @@ const CLOSED = Symbol('closed');
 const HALF_OPEN = Symbol('half-open');
 const PENDING_CLOSE = Symbol('pending-close');
 const FALLBACK_FUNCTION = Symbol('fallback');
+const FAILURE_CHECK_FUNCTION = Symbol('is-failure');
 const STATUS = Symbol('status');
 const NAME = Symbol('name');
 const GROUP = Symbol('group');
@@ -62,6 +63,7 @@ class CircuitBreaker extends EventEmitter {
     this[STATUS] = new Status(this.options);
     this[STATE] = CLOSED;
     this[FALLBACK_FUNCTION] = null;
+    this[FAILURE_CHECK_FUNCTION] = null;
     this[PENDING_CLOSE] = false;
     this[NAME] = options.name || action.name || nextName();
     this[GROUP] = options.group || this[NAME];
@@ -224,6 +226,24 @@ class CircuitBreaker extends EventEmitter {
       };
     }
     this[FALLBACK_FUNCTION] = fb;
+    return this;
+  }
+
+  /**
+   * Provide a failure check function to be called when the action throws/rejects an error.
+   * If this function returns true, then the error counts as a failure. Otherwise, the error
+   * is expected and should count as a success intead. Without calling this function, all
+   * errors are treated as failures that count towards opening circuit breaker.
+   *
+   * @param func {Function} a failure check function which returns a boolean.
+   *
+   * @returns {Boolean}
+   *
+   * @throws TypeError if `func` is supplied but not a function.
+   */
+  failureCheck (func) {
+    if (typeof func !== 'function') throw new TypeError('Failure check function must be a function');
+    this[FAILURE_CHECK_FUNCTION] = func;
     return this;
   }
 
@@ -423,18 +443,27 @@ function fallback (circuit, err, args) {
 }
 
 function fail (circuit, err, args, latency) {
-  /**
-   * Emitted when the circuit breaker action fails
-   * @event CircuitBreaker#failure
-   */
-  circuit.emit('failure', err, latency);
+  if (circuit[FAILURE_CHECK_FUNCTION] &&
+    !(circuit[FAILURE_CHECK_FUNCTION].call(circuit[FAILURE_CHECK_FUNCTION], err))) {
+    /**
+     * This error is not considered a failure
+     * @event CircuitBreaker#success
+     */
+    circuit.emit('success', err, latency);
+  } else {
+    /**
+     * Emitted when the circuit breaker action fails
+     * @event CircuitBreaker#failure
+     */
+    circuit.emit('failure', err, latency);
 
-  // check stats to see if the circuit should be opened
-  const stats = circuit.stats;
-  const errorRate = stats.failures / stats.fires * 100;
-  if (errorRate > circuit.options.errorThresholdPercentage ||
-    circuit.options.maxFailures >= stats.failures) {
-    circuit.open();
+    // check stats to see if the circuit should be opened
+    const stats = circuit.stats;
+    const errorRate = stats.failures / stats.fires * 100;
+    if (errorRate > circuit.options.errorThresholdPercentage ||
+      circuit.options.maxFailures >= stats.failures) {
+      circuit.open();
+    }
   }
 }
 
--- a/test/test.js
+++ b/test/test.js
@@ -732,6 +732,52 @@ test('CircuitBreaker semaphore rate limiting', (t) => {
   });
 });
 
+test('CircuitBreaker with failureCheck returns false', (t) => {
+  t.plan(3);
+  const breaker = cb(passFail, { errorThresholdPercentage: 10 });
+
+  breaker.failureCheck((e) => {
+    t.equals(e, 'Error: -1 is < 0', 'passed checked error');
+    return false;
+  });
+
+  breaker.fire(-1)
+    .then(() => t.fail('Tested function should throw an error'))
+    .catch((e) => t.equals(e, 'Error: -1 is < 0', 'caught expected error'))
+    .then(() => t.notOk(breaker.opened, 'Breaker should not open'))
+    .catch(t.end);
+});
+
+test('CircuitBreaker with failureCheck returns true', (t) => {
+  t.plan(3);
+  const breaker = cb(passFail, { errorThresholdPercentage: 10 });
+
+  breaker.failureCheck((e) => {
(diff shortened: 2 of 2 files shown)


lib/circuit.js
```diff
@@ -11,6 +11,7 @@ const CLOSED = Symbol('closed');
 const HALF_OPEN = Symbol('half-open');
 const PENDING_CLOSE = Symbol('pending-close');
 const FALLBACK_FUNCTION = Symbol('fallback');
+const FAILURE_CHECK_FUNCTION = Symbol('is-failure');
```
lance: Nit: the convention in `opossum` is that the `Symbol` name is the lowercase version of the `const` name, with `s/_/-/`.

lib/circuit.js
```diff
@@ -228,6 +230,24 @@ class CircuitBreaker extends EventEmitter {
   }
 
   /**
+   * Provide a failure check function to be called when the action throws/rejects an error.
```
lance: "when the action throws or rejects with an `Error`."

lib/circuit.js
```diff
@@ -423,18 +443,27 @@ function fallback (circuit, err, args) {
...
-  circuit.emit('failure', err, latency);
-
-  // check stats to see if the circuit should be opened
-  const stats = circuit.stats;
-  const errorRate = stats.failures / stats.fires * 100;
-  if (errorRate > circuit.options.errorThresholdPercentage ||
-    circuit.options.maxFailures >= stats.failures) {
-    circuit.open();
+  if (circuit[FAILURE_CHECK_FUNCTION] &&
+    !(circuit[FAILURE_CHECK_FUNCTION].call(circuit[FAILURE_CHECK_FUNCTION], err))) {
```
lance: As a general note, I'm not sure I like this naming convention. It makes the code a bit awkward by negating the `if` condition. Something more like `circuit[ERROR_OK]` maybe? That would make this code read a little better perhaps:

```js
if (circuit[ERROR_OK] && circuit[ERROR_OK].call(circuit[ERROR_OK], err)) {
  circuit.emit('success', err, latency);
}
```

Perhaps the class method could be `circuit.errorOkFilter( func )`.

test/test.js
```diff
@@ -732,6 +732,52 @@ test('CircuitBreaker semaphore rate limiting', (t) => {
...
+  t.plan(3);
+  const breaker = cb(passFail, { errorThresholdPercentage: 10 });
+
+  breaker.failureCheck((e) => {
+    t.equals(e, 'Error: -1 is < 0', 'passed checked error');
+    return false;
+  });
+
+  breaker.fire(-1)
+    .then(() => t.fail('Tested function should throw an error'))
```
lance: Nit: 'Tested function should return a rejected promise'

test/test.js
```diff
@@ -732,6 +732,52 @@ test('CircuitBreaker semaphore rate limiting', (t) => {
...
+  t.plan(3);
+  const breaker = cb(passFail, { errorThresholdPercentage: 10 });
+
+  breaker.failureCheck((e) => {
+    t.equals(e, 'Error: -1 is < 0', 'passed checked error');
+    return true;
+  });
+
+  breaker.fire(-1)
+    .then(() => t.fail('Tested function should throw an error'))
```
lance: It was only when I got to this test that I realized what's really happening here. In all cases, even if the error is determined to be ok, the circuit fails. This feature only prevents the statistic from being registered, but doesn't have an effect on the user code flow. 

I'm not sure about this. 

I think perhaps that `circuit.fire()` should return a resolved promise if the failure check function returns true. This is similar to how the `fallback` feature works. In the event of a failure, if a fallback function is available, the circuit resolves successfully. See: https://github.com/bucharest-gold/opossum/blob/master/test/test.js#L263-L266

lance:
I think I can see why you might want the action to fail with a rejected promise - doing it this way means you don't need to mess around with modifying code flow/control. What I'm concerned about is having this behavior affect the statistics and the behavior of the circuit in a non-intuitive way. Specifically, the behavior of the code is just as though there were an error, but a `'success'` event is emitted, even though the returned promise from `Circuit.fire()` is rejected. In addition, it modifies the true statistics of the behavior which can lead to non-intuitive reporting. 

Maybe there's a better way to go about it.

What about something like `Circuit#shouldOpen( func )` which takes a function that accepts a stats object, and the `Error` as a parameter? It returns a resolved Promise if the circuit should open, or a rejected one if not. The few places where `this.open()` is called in the circuit would be changed to call `_open(this, err)`.  This could also be a more flexible solution for user-controlled circuit behavior.

Maybe something like this?

```js

class CircuitBreaker {
  shouldOpen ( func ) {
    this[SHOULD_OPEN] = func;
  }
  // etc.
}

// outside of class scope
function _open (circuit, err) {
  if (circuit[SHOULD_OPEN])
    circuit[SHOULD_OPEN].call(circuit[SHOULD_OPEN], circuit.stats, err)
      .then(circuit.open)
      .catch(err => {
        // log error or shutdown or handle in some appropriate way
      });
  }
}

```

lance:
Maybe returning a promise in the `shouldOpen` function is overkill. A simple predicate function that returns `true` or `false` is clearer and simpler, and really the `shouldOpen` function shouldn't need to do anything asynchronous.

```js
// outside of class scope
function _open (circuit, err) {
  if (circuit[SHOULD_OPEN)) {
    return circuit[SHOULD_OPEN].call(circuit[SHOULD_OPEN], circuit.stats, err) && circuit.open();
  }
  return circuit.open();
}
```
@Permagate Do you think this would handle your use case?

lance:
@Permagate - just checking in to see if you have thoughts or feedback on my suggestions.

Permagate:
Thanks for the reminder! it was a bit hectic last week for me, so it slipped my mind to reply 😅 

Anyway, your suggestion of shouldOpen is quite different with what I have in mind. If I understand correctly, you'd like to keep the premise of where rejected promise should always be counted as `failure` and resolved promise as `success`, but having a `shouldOpen` function that will check if the failure should trip the circuit breaker or not.

However, the goal of this PR is to achieve what you mentioned:

> Specifically, the behavior of the code is just as though there were an error, but a 'success' event is emitted, even though the returned promise from Circuit.fire() is rejected.

While it might seem unintuitive, I think there is scenario where this is desirable. It is when the rejected error is expected, thus should not trip circuit breaker. For example, [axios](https://github.com/mzabriskie/axios) is a http library that rejects http requests when status code is not 2xx. Consider a function that tries to authenticate user:

```javascript
const breaker = require('opossum');
const axios = require('axios');

const wrappedPost = breaker(axios.post);

module.exports = () => {
  return wrappedPost.post('/login', {
    email: 'test@example.com',
    password: 'awdawdawd'
  });
};
```

Suppose if the user fails to authenticate (e.g. incorrect password) and status code is 401, the function will reject an error, and thus it will be counted as failure by circuit breaker. Given enough 401 error response, the circuit breaker will potentially trip, even though this error is expected. Ideally, circuit breaker should only trip in unexpected scenario which usually have 5xx error code. This is possible by adding `checkFailure` function like so:

```javascript
const breaker = require('opossum');
const axios = require('axios');

const wrappedPost = breaker(axios.post);
wrappedPost.checkFailure(err => {
  return err.respose.status >= 500;
});

module.exports = () => {
  return wrappedPost.post('/login', {
    email: 'test@example.com',
    password: 'awdawdawd'
  });
};
```

Now, even though the user fails to authenticate bazillion of times, it will never trip circuit breaker. With `shouldOpen`, if there are a lot of failed authentications, it will also not trip circuit breaker. However, because they are still counted as failures, just one 5xx error response is enough to trip circuit breaker. With `checkFailure`, multiple 5xx error responses are still required to trip.

Hope it makes sense! I'll fix the other feedbacks you mentioned above.

Permagate:
What I'd like to achieve is quite similar with `isFailure` in [levee](https://github.com/krakenjs/levee#isfailure).

lance:
@Permagate your example is helpful and illustrative. I think maybe you are misunderstanding my concern. I am ok with a `success` event being emitted if the failure check function returns `false`. My real concern is that the `Promise` returned from `fire()` is rejected even though the event is a `success`.

I took a look at the `levee` implementation, and this is the key difference between what you have here and what is happening in `levee`. In `levee`, if the failure check function returns false (e.g. there was no failure), then the user-provided callback is invoked as though there were no error. https://github.com/krakenjs/levee/blob/master/lib/breaker.js#L113-L119

I can see how this might be problematic in your example. You don't want a failed authentication to trip the circuit breaker, but at the same time, a failed authentication probably demands a different code path than a successful auth. However, it seems counter-intuitive to me that an invocation that results in a `success` event returns a rejected `Promise. In our case, to have parity with levee, the `Promise` returned from `fire()` should be resolved, not rejected. 

As I said when you first submitted this PR, I am ok with the idea in general and am grateful for the contribution. But the behavior of the code should be consistent, e.g. a `success` event always results in a resolved `Promise`.

lance:
@Permagate I'm going to close this for now. If you decide to pick it up again, it can be reopened.

Issue #109, opened 2017-11-10
Build error on Windows 10


helio-frota:
**Node.js Version:**
8.9.0
**Operating System:**
Windows 10
**Steps to Produce Error:**
`npm it`

```
$ npm it
npm WARN registry Unexpected warning for https://registry.npmjs.org/: Miscellaneous Warning EINTEGRITY: sha512-TWILfEnvO7I8mFe35d98F6T5fbLaEtbFTG/lxWvid8qDfFTxt19EBijWmB4j3+Hoh5TfHE2faWs73ua+EphuBA== integrity checksum failed when using sha512: wanted sha512-TWILfEnvO7I8mFe35d98F6T5fbLaEtbFTG/lxWvid8qDfFTxt19EBijWmB4j3+Hoh5TfHE2faWs73ua+EphuBA== but got sha1-9qn+xBzFCh3lD6M2A6tYCZH2Bo4=. (23892 bytes)
npm WARN registry Using stale package data from https://registry.npmjs.org/ due to a request error during revalidation.

> electron@1.6.11 postinstall C:\Users\hf\Desktop\dev\opossum\node_modules\electron
> node install.js

npm WARN prepublish-on-install As of npm@5, `prepublish` scripts are deprecated.
npm WARN prepublish-on-install Use `prepare` for build steps and `prepublishOnly` for upload-only.
npm WARN prepublish-on-install See the deprecation note in `npm help scripts` for more information.

> opossum@1.3.0 prepublish C:\Users\hf\Desktop\dev\opossum
> nsp check && npm run build:browser

(+) No known vulnerabilities found

> opossum@1.3.0 build:browser C:\Users\hf\Desktop\dev\opossum
> browserify index.js lib/*.js > dist/opossum.js

Error: Cannot find module 'C:\Users\hf\Desktop\dev\opossum\lib\*.js' from 'C:\Users\hf\Desktop\dev\opossum'
    at C:\Users\hf\Desktop\dev\opossum\node_modules\resolve\lib\async.js:55:21
    at load (C:\Users\hf\Desktop\dev\opossum\node_modules\resolve\lib\async.js:69:43)
    at onex (C:\Users\hf\Desktop\dev\opossum\node_modules\resolve\lib\async.js:92:31)
    at C:\Users\hf\Desktop\dev\opossum\node_modules\resolve\lib\async.js:22:47
    at FSReqWrap.oncomplete (fs.js:152:21)
npm ERR! code ELIFECYCLE
[...]

PR #112, opened 2017-11-23
fix: build on windows10 due browserify limitations

helio-frota:
Connects to #109


lance:
@helio-frota thanks for this fix. I think we should consider removing browserfy in the long run.

PR #120, opened 2017-12-12
Half-open state does not keep rejecting, and doesn't trigger re-try

jerryjvl:
Proposed fix for issues reported in https://github.com/bucharest-gold/opossum/issues/119

This fix addresses the fact that once the circuit goes "half-open" it lets all further requests through, not just the first one. And the follow-on issue that failure of the first request let through does not guarantee the circuit gets treated as "open" again.

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -270,7 +270,7 @@ class CircuitBreaker extends EventEmitter {
       this.emit('cacheMiss');
     }
 
-    if (this.opened && !this.pendingClose) {
+    if (!this.closed && !this.pendingClose) {
       /**
        * Emitted when the circuit breaker is open and failing fast
        * @event CircuitBreaker#reject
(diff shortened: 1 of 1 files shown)


lance:
@jerryjvl thanks for the contribution. That's definitely a race condition I did not consider. Would you be willing to write a test for this as well?

lance:
Connects to: https://github.com/bucharest-gold/opossum/issues/119

lance:
Landed in: https://github.com/bucharest-gold/opossum/commit/04df6f7b9a9b9e9ce672ea1665f1d95586f039a6

@jerryjvl thanks again for contributing this fix! I have added a test in https://github.com/bucharest-gold/opossum/commit/602a43c7f5097abbf091a07af9bf326a7b532718. I do not have a regular release cycle for this, so if you are need of a patch level release, let me know and I can push one out.

jerryjvl:
Thanks! I didn't get any messages before I checked in here, will have to check up my email settings!
Would have happily contributed a test case as well if I'd seen it earlier.

No rush, we're working off my fork for now, once this one has an official release we'll pick that up.

lance:
@jerryjvl s'all good - I pushed a 1.3.1 release with this fix last week: https://github.com/bucharest-gold/opossum/releases/tag/v1.3.1

Issue #125, opened 2017-12-19
[FEATURE REQUEST] Aggregate circuit stats


DavideCarvalho:
In order to see all the metrics of the circuits on Hystrix Dashboard, we need a way to aggregate all the stats of the circuits created in one stream.

This is just a feature request based on issue #124

PR #140, opened 2018-01-30
feat: allow multiple circuits to aggregate stats

lance:
If a user creates multiple circuits in the same process, they should aggregate
their stats in the Hystrix stream.

Fixes: https://github.com/bucharest-gold/opossum/issues/125
Ref: https://github.com/bucharest-gold/opossum/issues/124
Connects: #125


examples/hystrix/app.js
```diff
@@ -48,16 +48,8 @@ function fallback () {
...
-
 // setup our SSE endpoint
-app.use('/hystrix.stream', (request, response) => {
-  response.writeHead(200, {'Content-Type': 'text/event-stream', 'Cache-Control': 'no-cache', 'Connection': 'keep-alive'});
-  response.write('retry: 10000\n');
-  response.write('event: connecttime\n');
-
-  hystrixStats.getHystrixStream().pipe(response);
-});
+app.use('/hystrix.stream', require('./hystrix-stream.js')(circuitBreaker));
```
lholmquist: i don't think you need the `.js` extension here

index.js
```diff
@@ -45,6 +45,8 @@
    */
   circuitBreaker.promisify = require('./lib/promisify');
 
+  circuitBreaker.stats = require('./lib/hystrix-stats.js').stream;
```
lholmquist: same as other comment about the extension

PR #189, opened 2018-05-01
test: ensure all tests are run in the browser

### LINKED ISSUE #184: all tests are not run in the browser/headless environment
We are only running `test/test.js` and none of the other test files. We should be adding all of the tests to the browser test run.

lance:
Fixes #184


lholmquist:
not really sure about webpack configs and such,  but if the tests pass then 👍

Issue #196, opened 2018-05-23
Should fallback be applied when circuit is closed?


duartemendes:
This is not a bug per say, but I'm also not sure this is the correct behavior.

Shouldn't the fallback function only be applied when the circuit is open or half open?

I'm using this package to make http requests to several upstreams. And having the fallback being called when the fire action fails means that we lose the actual response from the upstream.

My suggestion is to invoke the fallback function only when the event 'reject' is also emitted. This will allow us to propagate the actual upstream response to above. Since we already have the response we might as well use it for debug and monitoring purposes.

hugoduraes:
@duartemendes Alternatively, what if the error was passed as a parameter to the fallback function? Would that be useful? Have a look at my PR #197

duartemendes:
Thanks @hugoduraes. I tried it out and it solves my problem.

@lance can you take a look at @hugoduraes PR?

lance:
@duartemendes it's important for the fallback function to be executed whether the circuit is closed, open or half open, otherwise there would simply be a failure. The solution provided by @hugoduraes is great, in my opinion. I will merge it and close this issue. Thanks both of you for contributing. Would you mind telling me what project you are using this in, if it's open source?

hugoduraes:
Thanks @lance. :)

PR #197, opened 2018-05-23
feat: pass error as parameter to fallback function

hugoduraes:
Possible solution for #196.


lance:
@hugoduraes thanks for the contribution!

Issue #198, opened 2018-05-25
Fallback is being called twice


duartemendes:
Considering that the circuit is closed and that the fallback function always returns a promise rejection: If the action triggered by the breaker is rejected and it took longer than the breaker timeout, the fallback function will be called twice. 

First time it's called because the breaker timeout is achieved. Second time it's called because the action was finished with a rejection. 

Due to the second call, we get the following error:
> (node:14582) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): StatusCodeError: 500 - {"statusCode":500,"message":"Server error"}
(node:14582) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

**Node.js Version:** 8.9.4

**Operating System:** Darwin DuarteMendes.local 17.5.0 Darwin Kernel Version 17.5.0: Mon Mar  5 22:24:32 PST 2018; root:xnu-4570.51.1~1/RELEASE_X86_64 x86_64

**Steps to Produce Error:**
```javascript
const circuitBreaker = require('opossum');

async function forceFallbackToBeExecutedTwiceOnTheSameExecution () {
    const breaker = circuitBreaker(timerBasedFunction, { timeout: 3500 });

    breaker.fallback((timer, err) => Promise.reject(err));
    breaker.on('fallback', () => console.log('Fallback was called!'));

    [0, 1, 2, 3, 4].forEach(async function (i) {
        try {
            const result = await breaker.fire(i * 1000);
            console.log(result);
        } catch (err) {
            console.log(err);
        }
    });
}

function timerBasedFunction (timer) {
    return new Promise((resolve, reject) => {
        setTimeout(() => {
            if (timer >= 4000) {
                return reject(`Error for the ${timer}ms timer`);
            }

            return resolve(`Result for the ${timer}ms timer`);
        }, timer);
    });
}

forceFallbackToBeExecutedTwiceOnTheSameExecution();
```

Result:
```
Result for the 0ms timer
Result for the 1000ms timer
Result for the 2000ms timer
Result for the 3000ms timer
Fallback was called!
{ Error: Timed out after 3500ms
    at Timeout.setTimeout (/Users/duartemendes/projects/testing-opossum/node_modules/opossum/lib/circuit.js:322:15)
    at ontimeout (timers.js:475:11)
    at tryOnTimeout (timers.js:310:5)
[...]

PR #201, opened 2018-05-27
fix: avoid calling fallback function twice (#198)

duartemendes:


duartemendes:
@lance test added.

lance:
@duartemendes thanks again for your contribution!

Issue #200, opened 2018-05-26
Optional timeout


duartemendes:
As a client I would like to optout the timeout feature.

Reason is that my action is a request, and this request already has a timeout specified. And I would rather receive the error thrown by the request than the opossum one.

For now, I have been setting the opossum timeout bigger than the request one, but this just doesn't feels right :)

lance:
Fixed in #202

PR #202, opened 2018-05-28
feat: optional timeout (#200)

duartemendes:


duartemendes:
I'm aware of the default value. On my project, I'm setting the timeout as false, so it's more clear and the default behaviour is still the same, so we aren't doing any breaking changes.

Not sure how I could test this, any tips?

lance:
@duartemendes it's non-optimal because of the fact that it will take > 1 second, but here is a stake in the ground.

```js
test('timeout is not called', t => {
  t.plan(1);
  // set up circuit breaker with timeout disabled
  const cb = circuitBreaker( { timeout: false } );
  cb.on(timedFunction(3000), 'timeout', t.fail);

  // 1000ms is the default timeout value
  // this test passes if we exceed this timeout value
  setTimeout(t.ok, 1500);
});
```

If you could also add a note to the [docs](https://github.com/bucharest-gold/opossum/blob/master/lib/circuit.js#L32) indicating this behavior, I would appreciate it. Thanks!

lance:
@duartemendes I've made these changes locally but can't push to your branch. Here is a diff: https://gist.github.com/lance/6a630d182cdad6ec8456bf3ee42afbcd

duartemendes:
@lance sorry for the delay, went for a little vacation.

I understand your test, however isn't the default timeout 10000ms? :)

I tried the test with timedFunction as 12000 and timeout as 15000 but tests started failing:

     Allows timeout to be optional

    ✔ function called

    ✖ plan != count
    ----------------
      operator: fail
      expected: 1
      actual:   2
      at: circuit.healthCheck._ (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:11:7)
      stack: |-


    ✖ .end() called twice
    ----------------------
      operator: fail
      at: circuit.healthCheck._ (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:12:7)
      stack: |-

    ✔ health-check-failed emitted

    ✖ plan != count
    ----------------
      operator: fail
      expected: 1
      actual:   2
      at: CircuitBreaker.circuit.on.e (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:21:7)
      stack: |-


    ✖ .end() called twice
    ----------------------
      operator: fail
      at: CircuitBreaker.circuit.on.e (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:22:7)
      stack: |-

    ✔ function called

    ✖ .end() called twice
    ----------------------
      operator: fail
      at: circuit.healthCheck._ (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:12:7)
      stack: |-

    ✔ health-check-failed emitted

    ✖ .end() called twice
    ----------------------
      operator: fail
      at: CircuitBreaker.circuit.on.e (/Users/duartemendes/tui/projects/opossum/test/health-check-test.js:22:7)
      stack: |-

    ✔ should be truthy

Updated the docs though.

lance:
@duartemendes OMG - an extra zero makes a lot of difference. Thanks for your help with this!

duartemendes:
No problem @lance! Please update package on npm :)

lance:
@duartemendes 1.7.0 is now available on npmjs.com and here https://github.com/bucharest-gold/opossum/releases/tag/v1.7.0

PR #203, opened 2018-05-31
Updated README file in JQuery example

rhoads-zach:
@lance @lholmquist let me know what you think

### DIFF
--- a/examples/jquery/README.md
+++ b/examples/jquery/README.md
@@ -1,11 +1,29 @@
 # JQuery Example
 
-This example exposes a simple service at the route `http://localhost:3000/flakeyService`. As the service receives requests, it gets slower and slower. Once it takes more than 1 second to respond, the service just returns a `423 (Locked)` error.
+This example exposes a simple service at the route `http://localhost:3000/flakeyService`. 
+For every request the service receives, the response time is increased.
+[The service returns a `423 (Locked)` error if the response time is above 1000ms.](https://github.com/bucharest-gold/opossum/blob/master/examples/jquery/index.js#L43-L46) This example also has a web frontend at `http://localhost:3000/` for interacting with the service.
 
-Start the server.
+To build and run this example:
 
-```sh
-$ npm start
-```
+1. From the root of the opossum repository, build the `opossum` module:
+  ```sh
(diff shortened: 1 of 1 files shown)


examples/jquery/README.md
```diff
@@ -1,11 +1,27 @@
...
+For every request the service receives, the response time is increased.
+[The service returns a `423 (Locked)` error if the response time is above 1000ms.](https://github.com/bucharest-gold/opossum/blob/master/examples/jquery/index.js#L43-L46) This example also has a web frontend at `http://localhost:3000/` for interacting with the service.
 
-Start the server.
+To build and run this example:
 
-```sh
-$ npm start
-```
+1. From the root of the project, build the `opossum` module:
```
lance: Maybe this should say, "from the root of the opossum repository", since there are two "projects" we are dealing with: the opossum module itself, and the example project.

examples/jquery/README.md
```diff
@@ -1,11 +1,27 @@
...
+  ```sh
+  $ npm install && npm run build:browser
```
lance: This is such a minor nit, so take it for whatever it's worth. But I often try to avoid using the "`&&`" shell syntax in examples like this, because it doesn't work the same with all shells. For example, fish shell, which I use would be `npm install ; and npm build:browser`. These could ostensibly be on two separate lines:

```sh
$ npm install
$ npm build:browser
```

But again... this is a nitpick.

examples/jquery/README.md
```diff
@@ -1,11 +1,27 @@
...
+  This builds the `opossum.js` file [used in this example](https://github.com/bucharest-gold/opossum/blob/master/examples/jquery/index.js#L21). You only need to do this once.
+
+1. Navigate to the `examples/jquery/` directory
+  ```sh
+  $ cd examples/jquery/
+  ```
+
+1. Start the server.
+  ```sh
+  $ npm install && npm start
```
lance: Same nit as above.

rhoads-zach:
thank you @lance for the feedback.

Updated.

lance:
Thanks for the help with this @rhoads-zach!

PR #218, opened 2018-07-20
feat: add options.allowWarmUp as a creation option

lance:
This feature allows the user to let the circuit warm up
before opening, even if every request is a failure or timeout.
The warmup duration is the value provided for
options.rollingCountTimeout.

Fixes: https://github.com/bucharest-gold/opossum/issues/217


helio-frota:
LGTM :+1:

PR #243, opened 2018-11-16
fix: error message for open breaker does not have error code.

arjun1995:
I create an Error with the existing message and add an error code to match the other type of errors. This also fixes issues with receiving only a string in fallbacks if the breaker is open.


lance:
Thanks for the contribution @arjun1995!

Issue #248, opened 2018-12-27
Memory leak. Inability to reset and/or gracefully destroy breakers.


mastermatt:
**Node.js Version:** v10.13.0
**Operating System:** Darwin Kernel Version 17.7.0 and 4.9.125-linuxkit
**Steps to Produce Error:** Call `new circuitBreaker(...)` more than 100 times in a single proc will produce the following log:
```
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 unpipe listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 drain listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 error listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 close listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 finish listeners added. Use emitter.setMaxListeners() to increase limit
```

This issue was alluded to in #181 and #245, and I understand that the expected usage of this lib is to implement a singleton pattern for one (or small number) of circuit breakers.

My issue is with testing apps or libs that use this lib. There currently doesn't seem to be a way to ensure that circuit breakers have a clean slate between each test. My current workaround is to remove the reference of the singleton breaker in my code, which has the effect of creating new breakers even though the old ones are still in memory and registered with the `hystrixStream Transform`.

I can think of a few different approaches to solve this and I don't think they are mutually exclusive. 
- Graceful shutdown of circuit breakers.
  Give the `CircuitBreaker` class a method that shuts itself off. The idea would be that any future calls to `fire` would fail and once any local refs to the instance were gone the GC could clear it. This would involve calling out to its `hystrixStats` and adding a method to the `HystrixStats` class that basically calls `_readableStream.unpipe(hystrixStream)`.
- Reset individual circuit breakers.
  Give the `CircuitBreaker` class a method that resets it back to a state that mimics a new instance. At a minimum, this would need to set the circuit to closed and update its `Status` instance with all new buckets for its window. Decisions would need to be made on whether this method would also empty the semaphore, enable a disabled circuit, and/or restart the warm up logic.

lance:
[...]

Issue #245, opened 2018-12-04
GC problem


HOTTIN:
**Node.js Version: V8.5.0**

**Operating System: centos 7**

**Steps to Produce Error:**
when i user opossum like this, i got a GC problem:
```js
class rqService extends app.Service {
    async request(url, opts) {
      opts = Object.assign(
        {
          timeout: [ '30s', '30s' ],
          dataType: 'json',
        },
        opts
      );
      const { ctx } = this;
      const breaker = new circuitBreaker(async function() {
        // create http request here
        return await ctx.curl(url, opts);
      }, breakerConfig);
      breaker.fallback(err => {
        ctx.logger.error(`[circuitBreaker]-fallback: ${url} got error`, err);
        return {
          code: '502',
          message: 'service not available',
          data: '',
        };
      });

      this.logger.info(
        'request external service: %s with params',
        url,
        opts && opts.data
      );
      const t1 = Date.now();
      const result = await breaker.fire().then(r => {
        const cost = Date.now() - t1;
        this.logger.debug('[%sms]response from [%s]: %j', cost, url, r);
        return r && r.data;
      });
      return result;
    }
  }
```
then i use jmeter to test my application,i got CPU profile like this:
![image](https://user-images.githubusercontent.com/12428754/49413911-66724300-f7ac-11e8-989e-baa762f476e3.png)
![image](https://user-images.githubusercontent.com/12428754/49416325-8f96d180-f7b4-11e8-98ba-6d5ee00248a9.png)

Issue #181, opened 2018-04-11
Performance issues


thonythony:
**Node.js Version:**
node v8.9.4
**Operating System:**
Linux 4.14.0-deepin2-amd64 #1 SMP PREEMPT Deepin 4.14.12-2 (2018-01-06) x86_64 GNU/Linux

**Steps to Produce Error:**
I have a component A that calls a component B using an http request.
When I benchmark this action using ab, without opossum, the average time per request is 50ms.
When I **wrap the http request into opossum** and I benchmark this action with ab, the average time per request **increases exponentially after each request** and errors are printed to console : 
```
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 unpipe listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 drain listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 error listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 close listeners added. Use emitter.setMaxListeners() to increase limit
(node:25619) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 finish listeners added. Use emitter.setMaxListeners() to increase limit
```

The ab command I use :
```
ab -n 10000 -c 100 http://localhost:3000/check1
```

Can you help me please ?

lance:
@thonythony do you experience this issue if you downgrade to version 1.3.1?

helio-frota:
@thonythony I could not reproduce.

Here a POC with the results:
https://github.com/helio-frota/opc

thonythony:
Hello,

@lance yes we tested with 1.3 and 1.4

@helio-frota thanks for your poc, I tested it and it works perfectly !

I see ours implementations differ. Unlike you, In my tests I instantiate a new circuit breaker instance for each request made. I think the memory leak and performance issues occur when a lot of circuit breaker are instantiated. 
We will change our code and implement a singleton service. I will keep you informed.

lance:
Thanks for the followup @thonthony  - I'm going to close this for now. If something else comes up related to it, please feel free to reopen or create a new issue.

dptoot:
[...]

PR #250, opened 2019-01-07
feat: add circuit.shutdown() to shut a circuit down

lance:
Shutting down a circuit will cause it to always return a
rejected promise with an `ESHUTDOWN` `Error.code` when
`circuit.fire()` is called, and the Hystrix stats stop
accumulating. Once a circuit has been shut down, it cannot
be reset or used again.

When using the `shutdown()` method on a circuit, all listeners
are cleaned up internally, preventing memory leaks.

Fixes: https://github.com/nodeshift/opossum/issues/248
Fixes: https://github.com/nodeshift/opossum/issues/181
Fixes: https://github.com/nodeshift/opossum/issues/245


lib/circuit.js
```diff
@@ -187,6 +187,25 @@ class CircuitBreaker extends EventEmitter {
...
+    this.disable();
+    this.removeAllListeners();
+    this.hystrixStats.shutdown();
+    this[HYSTRIX_STATS] = undefined;
+  }
+
+  /**
+   * Determines if the circuit has been shutdown.
+   */
+  isShutdown () {
```
mastermatt: A getter feels more appropriate given the rest of the attributes below are defined that way.

lib/circuit.js
```diff
@@ -187,6 +187,25 @@ class CircuitBreaker extends EventEmitter {
...
 
+  /**
+   * Shuts down this circuit breaker. All subsequent calls to the
+   * circuit will fail, returning a rejected promise.
+   */
+  shutdown () {
+    this.disable();
+    this.removeAllListeners();
+    this.hystrixStats.shutdown();
+    this[HYSTRIX_STATS] = undefined;
```
mastermatt: Unsetting the hystrix stats to denote the breaker is shut down feels odd, especially when there is already a concept of state within the class. `this[STATE] = SHUTDOWN;` comes across as more meaningful/explicit.

lib/hystrix-stats.js
```diff
@@ -11,12 +11,6 @@ const hystrixStream = new Transform({
   }
 });
 
-// TODO: This number is somewhat arbitrary. However, we need to allow
-// a potentially large number of listeners on this transform stream
-// because all circuits are connected to it. In an application with
-// a large number of circuits (or tests using circuits), if left to
-// the default MAX_LISTENERS, the user will see errors.
-hystrixStream.setMaxListeners(100);
```
mastermatt: I agree this should go, but some implementers may start seeing a warning. The default is 10, so anyone who has 11-100 instances in their app would start to see the log.

lib/hystrix-stats.js
```diff
@@ -41,34 +35,45 @@ hystrixStream.resume();
...
-      _readableStream.push(
-        Object.assign({},
-          {
-            name: circuit.name,
-            closed: circuit.closed,
-            group: circuit.group,
-            options: circuit.options
-          }, stats));
-    });
+      if (this._readableStream) {
```
mastermatt: `_readableStream` is never unset in `shutdown()` so I think this will always be true.
Maybe save the listener on the instance and add`circuit.status.removeListener('snapshot', this._ snapshotListener)` to `shutdown()`. You don't need to keep the bind if you have the readable stream as a local const too.

lib/hystrix-stats.js
lance: Good point. But I wonder if that's not so terrible. Typically, this might indicate a bug. For example, I have seen cases where people have created a new circuit every time the function needs to be called. It's a gross misunderstanding of how things should work, but leaving this value at 100 might prevent the user from ever knowing something is wrong.

lib/hystrix-stats.js
lance: I think it makes more sense to add a shutdown() method to Status which removes all listeners, and call it from shutdown() in the circuit. As for _readableStream not ever being unset, this is true. However in my tests I sometimes get the error TypeError: Cannot read property 'push' of undefined which I find quite odd. The only thing I can figure is that a snapshot event is being emitted before the constructor completes. I'm going to poke around on this a bit.

lib/hystrix-stats.js
mastermatt: Maybe add a bit in the README for those who get the warning and go searching for what it means. Something that outlines they're probably not properly using a singleton pattern, or if they do have a valid use case they can call `shutdown()` or set a high max listeners manually.

mastermatt:
@lance would it be possible to have a minor release for this?

lance:
@mastermatt I just published 1.10.0. Enjoy!

Issue #276, opened 2019-02-20
Circuit gets stuck in halfOpen and blocks all calls


brbradford:
I believe this line change has introduced a bug: **if (stats.fires < circuit.volumeThreshold) return;**
Steps to duplicate...
1. Set volumeThreshold to something.
2. Cause the circuit to open because of failures.
3. After the circuit goes to halfOpen, and after time has passed so that stats.fires is back below volumeThreshold...
4. Fire again with a call that will fail.
5. pendingClose gets set to false, this line causes an early return and keeps the circuit in halfOpen.
6. The circuit is now stuck in halfOpen forever and all calls get blocked.

It seems like you shouldn't return early here if the circuit is halfOpen.

Wafel:
You have saved my life. I have just met this issue on test environment and couldn't reproduce it on local env. Thanks.

brbradford:
@lance thanks for the quick turnaround!

lance:
@brbradford @Wafel version 1.10.1 including this fix has been published.

PR #277, opened 2019-02-22
fix: eliminate bug where circuit can get stuck in halfOpen state forever

lance:
Fixes: https://github.com/nodeshift/opossum/issues/276


lance:
@brbradford and @wafel I have added a test and fix for this and it should be good to go. Please give it a spin. I will try to land a new release next week at the latest.

Issue #262, opened 2019-02-07
A way to ignore certain type of errors


RidgeA:
Hello! 

I use the `request` module with the opossum to communicate between microservices.

In case if other microservice returns HTTP 4xx error, this is treated as an error and could lead to breaker opening. But in major cases, 4xx HTTP status is a valid and expected result.

So, I need somehow to tell opossum to ignore certain types of error and return them as is and don't treat them as an error.

Is it feasible?

lance:
All `opossum` does is proxy calls to the function it is provided. You could write a function that wraps your `request` calls, checking for acceptable responses and return a resolved promise when success or a rejected promise when failure.

RidgeA:
Yeah, I could do this.
But 4xx errors I want to pass through opossum and return this response as an error (the same as without opossum) but without changing internal circuit breaker's stats. The subsequent code expects to receive an error in this case.
Other errors (such as 5xx or network errors) should lead to breaker opening.

ComputerWolf:
I agree that a 4xx error does not necessarily mean that the service is unhealthy and should be configurable to not count as a failure, but it still makes sense to pass the error through as expected.

lance:
I really think this is in the domain of the client application and not something that `opossum` should be doing. It does not make sense to inspect the return value of all functions protected by a circuit breaker, determine if the function is a remote call receiving a 404 and then performing some special behavior. This is code that belongs in userland, not the module.

ComputerWolf:
@lance, while agree that any logic should happen outside of the module, right now it is all or nothing. There is no way to conditionally decide if a response should trip the circuit breaker and I would argue that there are some cases where a response should bypass the breaker. In those cases it is up to the module to provide the functionality to do so.

RidgeA:
I totally agree that `opossum` should not inspect the errors and make a decision on to count this as an error or not. The logic should belong to called code and depends on many factors and business logic.

Let's imagine the following example (pretty close to that I encountered with):

```
const request = require("request-promise-native")
const {StatusCodeError} = require("request-promise-native/errors")

//....

let result;

//....

try {
   result = await request(url);
   //....
} catch (e) {
   if (e instanceof StatusCodeError && e.statusCode === 404)  {
      // an expected outcome - just to return an empty collection
      result = [];
   } else {
      // could be any other 4xx error or any of 5xx errors or any network error
      // this in an unexpected case and the error should be re-thrown
      throw e;
   }
}
```

[...]

PR #281, opened 2019-03-05
feat: add errorFilter option to bypass incrementing failure stats

lance:
Refs: https://github.com/nodeshift/opossum/issues/262

~Still requires documentation.~

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,6 +51,12 @@ const defaults = {
  * can open. This is similar to `options.allowWarmUp` in that no matter how many
  * failures there are, if the number of requests within the statistical window
  * does not exceed this threshold, the circuit will remain closed. Default: 0
+ * @param options.errorFilter {Function} an optional function that will be
+ * called when the circuit's function fails (returns a rejected Promise). If
+ * this function returns truthy, the circuit's failure statistics will not be
+ * incremented. This is useful, for example, when you don't want HTTP 404 to
+ * trip the circuit, but still want to handle it as a failure case.
+
  * @return a {@link CircuitBreaker} instance
  */
 function circuitBreaker (action, options) {
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -72,6 +72,11 @@ Please use options.errorThresholdPercentage`;
  * can open. This is similar to `options.allowWarmUp` in that no matter how many
  * failures there are, if the number of requests within the statistical window
  * does not exceed this threshold, the circuit will remain closed. Default: 0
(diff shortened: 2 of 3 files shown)


lance:
@RidgeA @ComputerWolf @jahredhope please take a look at the requested feature as implemented and let me know what you think.

index.js
```diff
@@ -51,6 +51,12 @@ const defaults = {
  * can open. This is similar to `options.allowWarmUp` in that no matter how many
  * failures there are, if the number of requests within the statistical window
  * does not exceed this threshold, the circuit will remain closed. Default: 0
+ * @param options.errorFilter {Function} an optional function that will be
+ * called when the circuit 's function fails (returns a rejected Promise). If
```
jahredhope: Is the extra space before each `'s` in this paragraph intentional? 
```suggestion
 * called when the circuit's function fails (returns a rejected Promise). If
```

RidgeA:
@lance 
This is exactly what I expecting. Looking forward to a landing of this PR.

I have only one doubt - wouldn't be better to accept as an `errorHandler` promise/async function as well?
Personally I don't nee this at the moment, but it could be useful in some cases?...

ComputerWolf:
@lance, yes, thank you for taking the time to implement this!

index.js
lance: Good catch. My IDE did some strange reformatting when I copied/pasted. Thought I had caught it all.

lance:
@RidgeA I'm concerned about asynchronous execution here - not sure why exactly, but in my thinking, the `errorFilter` function should be simple and straightforward, only checking the result of a received error. I can imagine a scenario where you _might_ want to call a remote service to determine if the error should really be considered an error. But at first blush, that sounds overly complex and might indicate that the architecture should be reconsidered.

RidgeA:
@lance 
Totally agree, this is could be redundant.
Anyway, if someone will need this feature - they could open a new issue and make PR to enhance the `errorHandler`

lance:
@RidgeA @ComputerWolf @jahredhope this has been merged. It will be a few weeks before it lands in a release, but you should be able to use it in a package.json now directly from GitHub.

RidgeA:
@lance 
Thanks a lot!

RidgeA:
Hi @lance,
could you give any forecast when these changes will be in npm?
My company rules prohibit the use of public packages from git repos directly, only NPM...

Thanks!

RidgeA:
@lance, would appreciate, thank you!

lance:
@RidgeA I just published v1.11.0 which includes this change.

Issue #247, opened 2018-12-17
Add alternative metrics support


dijitali:
Hystrix appears to now be deprecated [as of 19th Nov](https://github.com/Netflix/Hystrix/commit/a7df971cbaddd8c5e976b3cc5f14013fe6ad00e6#diff-04c6e90faac2675aa89e2176d2eec7d8).

Are there any current plans to support alternative metrics functionality?  Personally a [Prometheus](https://github.com/prometheus/prometheus)-compatible integration would be ideal but perhaps a more agnostic solution would be more future-proof and adaptable.

**Node.js Version:**
N/A

**Operating System:**
N/A

**Steps to Produce Error:**
N/A

lance:
@dijitali yes - we are considering adding prometheus support.

PR #283, opened 2019-03-13
WIP: prometheus client integration

helio-frota:
This is not working for now. 
An weird behavior needs to be investigated when calling `require('prom-client')`.
I'm going to push the changes on this PR to share what is happening.

[Error here](https://travis-ci.org/nodeshift/opossum/jobs/505934958#L498)

Connects to: https://github.com/nodeshift/opossum/issues/247

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -3,6 +3,10 @@
 const EventEmitter = require('events');
 const Status = require('./status');
 const HystrixStats = require('./hystrix-stats');
+let PrometheusMetrics;
+if (!process.env.WEB) {
+  PrometheusMetrics = require('./prometheus-metrics');
+}
 const Semaphore = require('./semaphore');
 
 const STATE = Symbol('state');
@@ -16,6 +20,7 @@ const STATUS = Symbol('status');
 const NAME = Symbol('name');
 const GROUP = Symbol('group');
 const HYSTRIX_STATS = Symbol('hystrix-stats');
+const PROMETHEUS_METRICS = Symbol('prometheus-metrics');
 const CACHE = new WeakMap();
 const ENABLED = Symbol('Enabled');
 const WARMING_UP = Symbol('warming-up');
@@ -139,7 +144,7 @@ class CircuitBreaker extends EventEmitter {
     this.on('cacheMiss', increment('cacheMisses'));
     this.on('open', _ => this[STATUS].open());
     this.on('close', _ => this[STATUS].close());
-    this.on('semaphore-locked', increment('semaphoreRejections'));
+    this.on('semaphoreLocked', increment('semaphoreRejections'));
 
     /**
      * Emitted after `options.resetTimeout` has elapsed, allowing for
@@ -171,6 +176,9 @@ class CircuitBreaker extends EventEmitter {
 
     // Register with the hystrix stats listener
     this[HYSTRIX_STATS] = new HystrixStats(this);
+    if (PrometheusMetrics) {
+      this[PROMETHEUS_METRICS] = new PrometheusMetrics(this);
+    }
   }
 
   /**
@@ -220,6 +228,9 @@ class CircuitBreaker extends EventEmitter {
     this.removeAllListeners();
     this.status.shutdown();
     this.hystrixStats.shutdown();
+    if (PrometheusMetrics) {
+      this.metrics.clear();
+    }
     this[STATE] = SHUTDOWN;
   }
 
@@ -305,6 +316,14 @@ class CircuitBreaker extends EventEmitter {
   }
 
   /**
+   * Get the prometheus metrics for this circuit
+   * @type {PrometheusMetrics}
+   */
+  get metrics () {
+    return this[PROMETHEUS_METRICS];
+  }
+
+  /**
    * Gets whether the circuit is enabled or not
    * @type {Boolean}
    */
@@ -363,7 +382,7 @@ class CircuitBreaker extends EventEmitter {
    * @fires CircuitBreaker#reject
    * @fires CircuitBreaker#success
    * @fires CircuitBreaker#timeout
-   * @fires CircuitBreaker#semaphore-locked
+   * @fires CircuitBreaker#semaphoreLocked
    */
   fire () {
     if (this.isShutdown) {
@@ -488,10 +507,10 @@ class CircuitBreaker extends EventEmitter {
         /**
          * Emitted when the rate limit has been reached and there
          * are no more locks to be obtained.
-         * @event CircuitBreaker#semaphore-locked
+         * @event CircuitBreaker#semaphoreLocked
          * @type {Error}
          */
-        this.emit('semaphore-locked', err, latency);
+        this.emit('semaphoreLocked', err, latency);
         handleError(err, this, timeout, args, latency, resolve, reject);
       }
     });
@@ -522,7 +541,7 @@ class CircuitBreaker extends EventEmitter {
    *
    * @returns {void}
    *
-   * @fires CircuitBreaker#health-check-failed
+   * @fires CircuitBreaker#healthCheckFailed
    * @throws {TypeError} if `interval` is supplied but not a number
    */
   healthCheck (func, interval) {
@@ -539,10 +558,10 @@ class CircuitBreaker extends EventEmitter {
         /**
          * Emitted with the user-supplied health check function
          * returns a rejected promise.
-         * @event CircuitBreaker#health-check-failed
+         * @event CircuitBreaker#healthCheckFailed
          * @type {Error}
          */
-        this.emit('health-check-failed', e);
+        this.emit('healthCheckFailed', e);
         this.open();
       });
     };
--- a/lib/prometheus-metrics.js
+++ b/lib/prometheus-metrics.js
@@ -1,0 +1,58 @@
+'use strict';
+
+const client = require('prom-client');
+
+// The current tests has circuit names like:
+// 'circuit one' (with blank space) and others like
+// 3beb8f49-62c0-46e0-b458-dcd4a62d0f48.
+// So to avoid "Error: Invalid metric name" we are changing the
+// circuit name to pass the tests.
+// More details:
+// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+let i = 0;
+function normalizePrefix (prefixName) {
+  prefixName = prefixName.replace(/ /g, '_');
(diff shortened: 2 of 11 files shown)


helio-frota:
After comment the `require('prom-client')` the error is gone but of course the build is broken since we don't have the client. [here](https://travis-ci.org/nodeshift/opossum/jobs/505935964#L789)

Now I'm using ES6 class approach and not able to run `npm link`  to move forward.

I deleted my previous approach using functions but was totally wrong anyway.

For example if we create a file `foo.js` with the content:

```
const client = require('prom-client');
console.log(client);
```

The build passes

helio-frota:
Probably need to add/update webpack configuration since the error is common related.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+class PrometheusMetrics {
+  constructor (circuit) {
+    this.circuit = circuit;
+    // this.client = require('prom-client');
```
lance: Not clear why this isn't working - I think it should. Then, subsequently, I suggest that you initialize an actual client instance here instead of in `collect()`.

```suggestion
    this.client = require('prom-client');
    // Use the circuit breaker's name as a prefix
    // Keep a handle on the interval so we can clear it later
    this.interval = client.collectDefaultMetrics({ prefix: `${circuit.name}_` });
```

The `collect()` method below can probably just all be added to this constructor, since I don't think we want to expose `collect()` externally.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
```
lance: I think you should put the `require()` statement here:

```suggestion
const client = require('prom-client');
```

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+class PrometheusMetrics {
+  constructor (circuit) {
+    this.circuit = circuit;
+    // this.client = require('prom-client');
+    this.collect();
```
lance: Here, create an array of all the measurements we want to take.

```suggestion
    const openCounter = new client.Counter({
      name: 'circuit-open',
      help: 'A cumulative count of the times the circuit has opened'
    });

    const fireCounter = new client.Counter({
      name: 'circuit-open',
      help: 'A cumulative count of the times the circuit has fired'
    });
    this.counters = [ openCounter, fireCounter ]; // etc.
```

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  collect (prefix) {
+    this.collectDefaultMetrics = this.client.collectDefaultMetrics({ prefix: prefix, timeout: 5000 });
+    this.circuit.on('open', () => {
```
lance: I think this code should probably all be in the constructor, but in any case, we should be collecting metrics for circuit events, so I don't think `clear()` makes sense here (and below). Instead, we should be counting these events with a [counter](https://www.npmjs.com/package/prom-client#counter).

```suggestion
    this.circuit.on('open', () => {
    this.counters.forEach( counter => {
      counter.inc();
    });
```

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+      this.clear();
+    });
+    this.circuit.on('halfOpen', () => {
+      this.clear();
+    });
+  }
+
+  clear () {
+    clearInterval(this.client.collectDefaultMetrics());
```
lance: Use `this.interval` here. Otherwise, you're just starting a new thing that will start counting.

```suggestion
    clearInterval(this.interval);
```

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+
+  collect (prefix) {
+    this.collectDefaultMetrics = this.client.collectDefaultMetrics({ prefix: prefix, timeout: 5000 });
+    this.circuit.on('open', () => {
+      this.clear();
+    });
+    this.circuit.on('halfOpen', () => {
+      this.clear();
+    });
+  }
```
lance: We should probably add a snapshot listener that resets all of the metrics for each snapshot. We don't want counters to increase indefinitely. Instead, reset the counter for each snapshot bucket.

```suggestion
  this.circuit.status.on('snapshot', _ => this.counters.forEach( counter => counter.reset() );
```

lib/circuit.js
```diff
@@ -271,6 +275,10 @@ class CircuitBreaker extends EventEmitter {
     return this[HYSTRIX_STATS];
   }
 
+  get prometheusMetrics () {
```
lance: We could probably just call this `metrics()` since if the target changes (e.g. no longer using Prometheus) we don't have to change the name.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+  clear () {
+    clearInterval(this.client.collectDefaultMetrics());
+    this.client.register.clear();
+  }
+
+  getMetrics () {
+    if (this.client.register && this.client.register.metrics) {
+      return this.client.register.metrics();
+    }
+    return "Can't get prometheus metrics.";
```
lance: I wonder if this shouldn't be a thrown `Error`, since it should never really happen.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  getClient () {
```
lance: This could just be a `get` property on the object. So instead of calling `metrics.getClient()` it becomes `metrics.client` as a read only property.

```suggestion
  get client () {
```

helio-frota:
@lance Thanks for the suggestions and review. :+1:

lib/prometheus-metrics.js
helio-frota: Probably yes, but the fact is the when we `require('prom-client')`  it automatically starts to collect the metrics, and I thought that was the reason for the errors. But seems the errors are related to webpack config which still under investigation.

lance: >  when we require('prom-client') it automatically starts to collect the metrics

I don't think this is actually the case. The metrics don't start being collected until `collectDefaultMetrics()` is called, or you provide a custom metric to the module.

helio-frota:
Now able to run the build locally, let's see what happen on CI.
After struggling for a good time trying to create a solution with multiple targets I decided to try it using two separated webpack configuration files.

helio-frota:
Build failed with tests but at least a progress.

helio-frota:
Using the `webpack.DefinePlugin` + ignoring the file `prometheus-metrics.js` we are able to add checks on the code to avoid the usage

```js
new webpack.DefinePlugin({
  WEB: JSON.stringify(true)
})
```

```js
if (!WEB) {
  const PrometheusMetrics = require('./prometheus-metrics');
}
const Semaphore = require('./semaphore');
```

And to add that checks on every part of the project that uses prometheus-metrics file. Using this we are able to run the browser tests but besides the ugly checks on the code we get with error on the node test:

> showing the browser test finished and starting the node test with failures:

```console
# CircuitBreaker semaphore rate limiting
ok 209 Semaphore rejection status incremented
ok 210 Semaphore was locked
# CircuitBreaker default capacity
ok 211 should be equal

1..211
# tests 211
# pass  211

# ok

✔ ~/Desktop/opossum [prometheus-client L|✚ 2] 
17:23 $ npm run test

> opossum@1.11.0 pretest /home/hf/Desktop/opossum
> npm run lint && npm run build:browser


> opossum@1.11.0 lint /home/hf/Desktop/opossum
> semistandard examples/*/*.js test/*.js index.js lib/*.js

semistandard: Semicolons For All! (https://github.com/Flet/semistandard)
  /home/hf/Desktop/opossum/lib/circuit.js:6:6: 'WEB' is not defined.
  /home/hf/Desktop/opossum/lib/circuit.js:7:9: 'PrometheusMetrics' is assigned a value but never used.
  /home/hf/Desktop/opossum/lib/circuit.js:164:10: 'WEB' is not defined.
  /home/hf/Desktop/opossum/lib/circuit.js:165:38: 'PrometheusMetrics' is not defined.
  /home/hf/Desktop/opossum/lib/circuit.js:213:10: 'WEB' is not defined.
  /home/hf/Desktop/opossum/lib/circuit.js:285:10: 'WEB' is not defined.
```

helio-frota:
While this is not an elegant solution IMO. It was able to run both browser and node tests locally. Let's see what happen on CI.

helio-frota:
Things I think need to be revisited/review/agreed/improved:

* Webpack configuration approach on build:node and build:browser 
* Probably to install opossum via github instead to release canary version for every change
* To use (a fork)  the [example](https://github.com/helio-frota/express-prometheus-minishift) to verify the results
* Currently we can see the metrics `circuit_open` and `circuit_fired` inside prometheus running on minishift but for some reason they are not increasing (on the graph) the values after run the bench test.



The current dev workflow for this:

1. make changes on opossum
2. release canary version
3. deploy example app 
4. run the bench
5. verify the results on prometheus web interface running on minishift (setup/config based on the example app)

helio-frota:
```console
                      === npm audit security report ===                        
                                                                                
┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ marked                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=0.6.2                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ jsdoc [dev]                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ jsdoc > marked                                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/812                             │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

@lance can we remove jsdoc? One of the deps has vulnerability.

lance:
@helio-frota we already have an issue for this https://github.com/nodeshift/opossum/issues/289

lance:
Closing this. Duplicated in https://github.com/nodeshift/opossum/pull/313

PR #313, opened 2019-04-30
feat: prometheus client integration

lance:
Adds Prometheus metrics to the circuit. As a result,
of limitiations with names in Prometheus, this also
normalizes all circuit event names to camelCase.

Fixes: https://github.com/nodeshift/opossum/issues/247

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -3,6 +3,10 @@
 const EventEmitter = require('events');
 const Status = require('./status');
 const HystrixStats = require('./hystrix-stats');
+let PrometheusMetrics;
+if (!process.env.WEB) {
+  PrometheusMetrics = require('./prometheus-metrics');
+}
 const Semaphore = require('./semaphore');
 
 const STATE = Symbol('state');
@@ -16,6 +20,7 @@ const STATUS = Symbol('status');
 const NAME = Symbol('name');
 const GROUP = Symbol('group');
 const HYSTRIX_STATS = Symbol('hystrix-stats');
+const PROMETHEUS_METRICS = Symbol('prometheus-metrics');
 const CACHE = new WeakMap();
 const ENABLED = Symbol('Enabled');
 const WARMING_UP = Symbol('warming-up');
@@ -146,7 +151,7 @@ class CircuitBreaker extends EventEmitter {
     this.on('cacheMiss', increment('cacheMisses'));
     this.on('open', _ => this[STATUS].open());
     this.on('close', _ => this[STATUS].close());
-    this.on('semaphore-locked', increment('semaphoreRejections'));
+    this.on('semaphoreLocked', increment('semaphoreRejections'));
 
     /**
      * Emitted after `options.resetTimeout` has elapsed, allowing for
@@ -178,6 +183,9 @@ class CircuitBreaker extends EventEmitter {
 
     // Register with the hystrix stats listener
     this[HYSTRIX_STATS] = new HystrixStats(this);
+    if (PrometheusMetrics) {
+      this[PROMETHEUS_METRICS] = new PrometheusMetrics(this);
+    }
   }
 
   /**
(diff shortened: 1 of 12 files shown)


lib/circuit.js
```diff
@@ -220,6 +228,9 @@ class CircuitBreaker extends EventEmitter {
     this.removeAllListeners();
     this.status.shutdown();
     this.hystrixStats.shutdown();
+    if (PrometheusMetrics && this.metrics) {
```
lance: I think this can be simpler.
```suggestion
    if (this.metrics) {
```
If `PrometheusMetrics` is `undefined`, this always will be as well.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,57 @@
...
+// The current tests has circuit names like:
+// 'circuit one' (with blank space) and others like
+// 3beb8f49-62c0-46e0-b458-dcd4a62d0f48.
+// So to avoid "Error: Invalid metric name" we are changing the
+// circuit name to pass the tests.
+// More details:
+// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
+function normalizePrefix (prefixName) {
+  prefixName = prefixName.replace(/ /g, '_');
+  prefixName = prefixName.replace(/-/g, '_');
```
lance: This can be written as a single statement instead of two. And instead of checking for a digit below, we can just always prefix with a string. So these two statements just become:

```suggestion
  return `circuit_${prefixName.replace(/[ |-]/g, '_')}_`;
```
And the following four lines can be removed.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,57 @@
...
+  }
+  return `${prefixName}_`;
+}
+
+class PrometheusMetrics {
+  constructor (circuit) {
+    this.circuit = circuit;
+    this._client = client;
+    this.counters = [];
+    const prefix = normalizePrefix(`${this.circuit.name}_`);
```
lance: Seems like you don't need the `_` at the end here, do you? We're already adding it in the `normalizePrefix()` function.

lib/prometheus-metrics.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.interval = this._client
+      .collectDefaultMetrics({ prefix, timeout: 5000 });
+
+    for (let eventName of this.circuit.eventNames()) {
+      const counter = new this._client.Counter({
+        name: `${prefix}circuit_${eventName}`,
```
lance: If you incorporate the suggestion above in `normalizePrefix()`, this would just become:
```suggestion
        name: `${prefix}${eventName}`,
```

lib/circuit.js
```diff
@@ -171,6 +176,9 @@ class CircuitBreaker extends EventEmitter {
 
     // Register with the hystrix stats listener
     this[HYSTRIX_STATS] = new HystrixStats(this);
+    if (PrometheusMetrics && options.usePrometheus) {
```
lance: What happens if `options.usePrometheus` is `true`, but `PrometheusMetrics` is undefined? Obviously, this conditional check will be false, but should we emit an error in that case?

lib/circuit.js
helio-frota: That's right. But I'm not sure/tested what happen with the webpack config

lib/prometheus-metrics.js
helio-frota: Yup. The idea was to put to work first, then improve later.

lib/prometheus-metrics.js
helio-frota: I keep that only to solve the issue with the tests.

lib/circuit.js
helio-frota: I think that part is more related to webpack config like the other situation commented above.
Prometheus is now enabled on two situations:
1) Web vs Node environment (webpack)
2) User defined option ( `usePrometheus` )

But yes, is good to emit an error in case the user try to use prometheus without `usePrometheus` option :+1:

PR #326, opened 2019-06-10
Add original function parameters to the failure and timeout events

### LINKED ISSUE #324: Passing function parameters to events
This is more of a question than an issue. I like to use the parameters that were passed in the action function and do something in the event.  For instance, let's say `x` and `y` are passed in `asyncFunctionThatCouldFail` function and I like to print out the `x` and `y` on timeout event.  How can I do that?

```
const circuitBreaker = require('opossum');

function asyncFunctionThatCouldFail (x, y) {
  return new Promise((resolve, reject) => {
    // Do something, maybe on the network or a disk
  });
}

const options = {
  timeout: 3000, // If our function takes longer than 3 seconds, trigger a failure
  errorThresholdPercentage: 50, // When 50% of requests fail, trip the circuit
  resetTimeout: 30000 // After 30 seconds, try again.
};
const breaker = circuitBreaker(asyncFunctionThatCouldFail, options);

breaker.fire(params);
circuit.on('timeout',
  () => console.log(`x equal ${x} and y equal ${y}`)); // the parameters of asyncFunctionThatCouldFail
```

lholmquist:
This adds the original functions parameters to the timeout and failure events as the third argument.

The third argument will be an array of values

fixes #324

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -808,6 +808,24 @@ test('Circuit Breaker timeout event emits latency', t => {
   breaker.fire(-1).catch(noop);
 });
 
+test('Circuit Breaker timeout event emits function parameters', t => {
+  t.plan(6);
+  const breaker = circuit(slowFunction, { timeout: 10 });
+
+  breaker.on('timeout', (result, latencyTime, args) => {
+    t.ok(args, 'third argument is the function args');
+    t.equal(Array.isArray(args), true, 'The args parameter is an array');
+    t.equal(args[0], -1, 'this is the first argument');
+    t.equal(args[1].arg1, 'arg1', 'this is the second argument object');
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -795,6 +795,23 @@ test('Circuit Breaker failure event emits latency', t => {
   breaker.fire(-1).catch(noop);
 });
 
+test('Circuit Breaker failure event emits function parameters', t => {
+  t.plan(6);
+  const breaker = circuit(passFail);
+  breaker.on('failure', (result, latencyTime, args) => {
+    t.ok(args, 'third argument is the latency');
```
lance:
```suggestion
    t.ok(args, 'third argument is the function args');
```

lholmquist: good ole copy/paste

PR #332, opened 2019-06-14
feat: add prometheusRegistry option

SerayaEryn:
At the moment it is only possible to add prometheus metrics to the global registry of [prom-client](https://github.com/siimon/prom-client).

This adds a `prometheusRegistry` option, that allows to provide a custom, non-global registry.
If a registry is being provided the default metrics will not be added to avoid having duplicate default metrics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,6 +259,27 @@ app.use('/metrics', (req, res) => {
 });
 ```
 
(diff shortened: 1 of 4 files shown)


helio-frota:
LGTM

PR #346, opened 2019-06-27
chore: docs updated with the correct version number on release

### LINKED ISSUE #345: build: add prerelease task that builds docs
Currently, it's easy to forget that a new release should have new docs published. By adding a `prerelease` task that runs `npm run build:docs` we don't have to remember to do this.

lholmquist:
…a release

* Adding the postbump release script allows the docs to be generated after the package.json version bump
* Adding the precommit release script allows those newly generated docs to be part of the release
* the -a flag makes sure to commit generated files

fixes #345

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,15 @@
     "coverage": "nyc report && if [ $CODACY_PROJECT_TOKEN'' != '' ] ; then nyc report --reporter=text-lcov | codacy-coverage; fi",
     "ci": "npm run build && npm run test && npm run test:headless && npm run coverage",
     "prerelease": "npm run ci",
-    "release": "standard-version -s",
+    "release": "standard-version -s -a",
     "lint": "standardx test/*.js index.js lib/*.js test/*/*.js",
     "clean": "rm -rf node_modules dist/*.js test/browser/webpack-test.js"
+  },
+  "standard-version": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -22,10 +22,16 @@
     "coverage": "nyc report && if [ $CODACY_PROJECT_TOKEN'' != '' ] ; then nyc report --reporter=text-lcov | codacy-coverage; fi",
     "ci": "npm run build && npm run test && npm run test:headless && npm run coverage",
     "prerelease": "npm run ci",
-    "release": "standard-version -s",
+    "release": "standard-version -s -a",
```
lance: I personally don't like to use `-a` because I often have scratch files in my project folder that would then mistakenly be added to the repo. Doesn't the `precommit` script for `standard-version` below add the correct files anyway? Seems like maybe this isn't necessary?

lholmquist: I will need to test,  but i don't think it will commit those scratch files, since according to the docs
`You will need to stage the artifacts you want to commit,`. 

i'll test to confirm though

lholmquist: Just tested, and the `-a` flag will only commit things that are already staged.  So if there are working files that havent yet been staged or are not yet under source control, those won't be included in the release commit

Issue #347, opened 2019-06-27
Metrics and Stats need to be cleaned up and refactored


lance:
The code that manages and exposes both Prometheus metrics and Hystrix stats is kind of all over the place and a bit confusing. I think this could be cleaned up quite a bit, and perhaps even refactored so that the metrics/statistics reporting can be provided via plugins. I.e. `opossum` has its own base set of metrics that it provides in its own chosen format. All other statistics derive from these (except for low level Node.js metrics provided by `prom-client`). It would be really cool if you could just write a plugin that accepted a stream of metrics - or alternatively listened for events. The plugin could then do whatever it wanted with the incoming stats: format it for Prometheus or Hystrix, log it to a database or key/value store, monitor for anomolies and send alerts. 

Also, by providing these as plugins, then the one dependency the project took on when moving to Prometheus (`prom-client`), could be extracted to live only within the `opossum-prometheus` plugin.

I don't think this would be terribly difficult to implement but would be a pretty big change in the surface API, therefore probably another major version bump.

Refs: https://github.com/nodeshift/opossum/issues/343
Refs: https://github.com/nodeshift/opossum/issues/342
Refs: https://github.com/nodeshift/opossum/issues/331

Issue #342, opened 2019-06-23
Store metrics for browser implementation


Sumeshkumar:
Store metrics for browser implementation so that its does not get reset while refresh the browser.

Use case: 

When browser refresh the metrics are getting reset currently. Can think of something like localStorage to store these values so that it does not reset.

This issue has been created as per the discussion on https://github.com/nodeshift/opossum/issues/333.

lance:
@Sumeshkumar thank you for submitting this issue. I have given it some thought and played around a bit locally, and I think this is actually best left to the user.  It should be simple enough to add something like this to user code. Here, for example, is a very simple implementation for session storage of circuit metrics using [`opossum-prometheus`](https://github.com/nodeshift/opossum-prometheus).

```js
const circuit = new CircuitBreaker(protectedFunc);
const metrics = new PrometheusMetrics(circuit);

circuit.status.on('snapshot', _ => {
  sessionStorage.setItem('metrics', prometheus.metrics);
});
```

I am going to go ahead and close this issue for now.

PR #350, opened 2019-07-01
feat: refactor Prometheus and Hystrix metrics

lance:
Instead of continuing to add different metrics modules to the
circuit breaker API, we can extract the `PrometheusMetrics` and
`HystrixStats` classes as their own modules. All they need is a
handle to the circuit itself so they can listen for events and/or
monitor the circuit stats. This allows us to continue maintaining
things like Hystrics that are deprecated upstream, without polluting
our circuit breaker implementation.

Fixes: https://github.com/nodeshift/opossum/issues/347
Fixes: https://github.com/nodeshift/opossum/issues/343
Related: https://github.com/nodeshift/opossum/issues/342


lance:
This is a work in progress. There are a couple of things to note about this change.

* It does not modify the API to provide a plugin interface. Instead, it removes API specific functions and properties related to any specific third party metrics tool - e.g. Prometheus and Hystrix.
* Prometheus and Hystrix metrics instances do not have to conform to any specific interface. It is up to the author, but we probably want to document some "good practices" - e.g. accept an array of circuits in the constructor.
* I believe that the functionality of the metrics implementations remains the same with one exception. ** Previously, the default Node.js metrics that `prom-client` gathers by default is now prefixed with `opossum_` instead of `circuit_${circuit.name}_`. This probably makes more sense anyway, because the Node.js metrics maintained by `prom-client` are not specific to any individual circuit, but rather to the entire process which may have multiple circuits.
* Before landing - or at least before a new version is released with this capability - we should probably extract the `PrometheusMetrics` and `HystrixStats` modules into an `opossum-metrics` repository and have these be maintained and versioned independent of the `opossum`.

/cc @ttessarolo @Sumeshkumar

ttessarolo:
@lance Just to share my view. 

I've recently developed a Microservices Framework for NodeJs (to be open sourced when test and docs will be done...) thus I've seen in action how powerful are two concepts:

- pluggability
- hooks

Thus to give you just a raw example on what I see for opossum (that I use in my framework):

```
const circuitBreaker = require("opossum");

const breaker = circuitBreaker(hanlderFunc, breakerOpts);
breaker.plug(require("opossum-histryx-plugin"), {...options...});
breaker.plug(require("opossum-prometheus-plugin"), {...options...});
breaker.plug(require("opossum-redis-cache-plugin"), {...options...});

// Then you have hooks (existing hooks for circuit status are still there)
breaker.on("fallback", fallbackFunc);
breaker.on("stats-emit"); // opossum-histryx-plugin + opossum-prometheus-plugin are registered to this hook
breaker.on("beforeCache", beforeCacheFunc); // chek if to cache --> return false === noCache
breaker.on("cache"); // opossum-redis-cache-plugin is registered to this hook
breaker.on("afterCache", beforeCacheFunc); // verify cache

// Start the Circuit --> activate plugings
breaker.exec();
```

The only doubt is if you really need to let plug-in to be registered (and configured) for every breaker instance or, instead, if you just want to provide a generi plugin registration for every breaker instace:

```
const circuitBreaker = require("opossum");

circuitBreaker.plug(require("opossum-histryx-plugin"), {...options...});
circuitBreaker.plug(require("opossum-prometheus-plugin"), {...options...});
circuitBreaker.plug(require("opossum-redis-cache-plugin"), {...options...});
``

The first option is more powerful, but maybe not so usefull.

That's very quickly my view.
As I said you can use npm Avvio to handle plug-in logic, and let plug-in register to hooks, cache handling included.

**Comment by lholmquist** [comment]
> Before landing - or at least before a new version is released with this capability - we should probably extract the PrometheusMetrics and HystrixStats modules into an opossum-metrics repository and have these be maintained and versioned independent of the opossum.

Are you thinking these would be in the same repo(published as 1 module) or should be separately published modules in their own repos.  I think i like the second option

**Comment by helio-frota** [comment]
> Before landing - or at least before a new version is released with this capability - we should probably extract the PrometheusMetrics and HystrixStats modules into an opossum-metrics repository and have these be maintained and versioned independent of the opossum.

Sounds good.

> Are you thinking these would be in the same repo(published as 1 module) or should be separately published modules in their own repos. I think i like the second option

+1

**Comment by lance** [comment]
@ttessarolo thanks for providing that insight. What I am having a hard time understanding is why the circuit needs to have any awareness of these plugins. The stats plugins such as Prometheus and Hystrix have everything they need by just listening for the `snapshot` event. What kind of behavior do you see the circuit actually taking care of other than just emitting the stats? What is the specific, tangible benefit of doing it the way you suggest?

**Comment by ttessarolo** [comment]
> What kind of behavior do you see the circuit actually taking care of other than just emitting the stats? What is the specific, tangible benefit of doing it the way you suggest?

Well, not an evident so clear tangible benefit. But for instance if you want to let devs implement a cache plugin (eg. using Redis) you want to be sure that the plugin is up-and-running before invoke it.
Plus we can share the plugin functionalities inside the contest of the handler function the eg. you can reference to redis inside the handler using `this.redis.set()` wher `this` is opossum itself.

And definitely opening opossum to plugin wide the development to scenarios not yet imagined.

File: lib/plugins/hystrix/hystrix-stats.js [hunk 1]
```diff
@@ -32,24 +32,26 @@ hystrixStream.resume();
  * @see CircuitBreaker#hystrixStats
  */
 class HystrixStats {
-  constructor (circuit) {
+  constructor (circuits) {
```
mcollina: I would add an API `add(circuit)` as well.

lholmquist:
umm,  something crazy happened with the commit history

lance:
@lholmquist @helio-frota @danbev I have moved the prometheus and hystrix metrics into their own modules in the @nodeshift org. Can y'all give this one more looksee? I'm eager to merge it and get this over with! :)

lance:
@ttessarolo I've decided to go ahead and keep this simpler for now and avoid the plugin pattern. I think that most or all of what you are asking for can be taken care of in user code. Additionally (though I have not tested it) doing it this way means that these metrics can be used on older versions of opossum, since no plugin knowledge is required.

lance:
@mcollina yes I will do those things for 4.0  but to keep it clean I would like to do these as separate independent pull requests.

mcollina:
Awesome, thanks!

LGTM

PR #369, opened 2019-08-23
fix: circuit breaker global should be capital.

lholmquist:
BREAKING CHANGE: since the circuit breaker is now a constructor, the web version should also be capitalized as CircuitBreaker. This is a breaking change since it is still lower case

@lance we really don't need to do this, since it is more a semantic thing,  but it would keep the usage consistent between browser and node

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -20,7 +20,6 @@ configs.push({
     extensions: ['*', '.js']
   },
   plugins: [
-    new webpack.IgnorePlugin(/prom-client/),
     new webpack.DefinePlugin({
       'process.env': {
         WEB: JSON.stringify('web')
@@ -34,13 +33,13 @@ function generateConfig (name) {
   const config = {
     mode,
     entry: {
-      circuitBreaker: './index.js'
+      CircuitBreaker: './index.js'
     },
     output: {
       path: path.resolve(__dirname, '..', 'dist'),
       filename: `${name}.js`,
       sourceMapFilename: `${name}.map`,
-      library: 'circuitBreaker',
+      library: 'CircuitBreaker',
       libraryTarget: 'umd'
     },
     node: {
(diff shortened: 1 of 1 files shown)


lance:
Good catch - I think maybe we can do this without `semver-major` by making the change additive instead of a strict replacement.

config/webpack.config.js
```diff
@@ -34,23 +33,22 @@ function generateConfig (name) {
...
     },
     node: {
       process: true,
       console: true
     },
     plugins: [
-      new webpack.IgnorePlugin(/prom-client/),
       new webpack.ProvidePlugin({
-        'circuitBreaker': 'opossum'
+        'CircuitBreaker': 'opossum'
```
lance:
```suggestion
        'CircuitBreaker': 'opossum',
        'circuitBreaker': 'opossum'
```

Not 100% sure, but I think this will keep backwards compatibility.

lholmquist: If that works, then that would be a good way to do it

lholmquist: @lance anymore thoughts on this?

I would like to keep things consistent, especially for the examples,  but it is not totally needed

lance: Sorry - I looked at this last week and just really couldn't decide. I think it will need to be at least a minor version bump (and probably should be major). I am inclined to keep it as-is to avoid having such a small change cause a big version bump.

lholmquist: that sounds reasonable.   I will keep this open though, so when we do have a bigger change that will require a major change, maybe that is when we add this

lholmquist: @lance in a recent commit we dropped node 8 support, 9145c9263118dc0ccbe5d2621373e5d0f801ed60 , so i was going to do a major release for it.  I'm thinking about getting this added, if we are going to do a major release.  It might make sense to do now.  any thoughts?

I'm still not totally sure about it though,  since it is just a coding convention

lance: Fine with me.

Issue #349, opened 2019-06-29
Updating action's context


nodify-at:
## Node.js Version:
v8.15.0
## Operating System:
Windows 10
## Steps to Produce Error:
Steps to reproduce is little bit complex. Let me try to explain the case:

We have an application based on Express und using a Dependency Injection Framework. Each Service Object has a web request scope.

For one of our service call, we need to use a circuit breaker. Each require/call of circuit breaker creates a new circuit break. But in this case we need a global circuit breaker to store failed counts etc.. Therefore we check globally whether a circuit breaker has been initialised for this function, if 
* yes: Don't create a new circuit break just call fire function
* no: Create a new circuit break and then call the function

We use a decorator to wrap this service before calling it and wrapping against by circuit breaker.

Problem: 
We need rebind the function to "this" to implemented class so we can call private functions and fields etc. 
As "this" can be bound only once, all other request will use the "this" scope of the first call.

```js
    const sourceFunction = descriptor.value; // save a reference to the source method
    const breaker;

    descriptor.value = async function(...args: any[]) {
        if (!breaker) {
            breaker = circuitBreaker(sourceFunction.bind(this), options);
        }
        return breaker.circuitBreaker.fire(args);
    };
```

```js
        if (!breaker) {
            breaker = circuitBreaker(sourceFunction.bind(this), options);
        }
        return breaker.circuitBreaker.fire(args);
```
This causes the issue, because we can bind "this" only once. Even the arrow functions don't help in this case. 

**Possible solution:**
I think we can solve this issue if we can send the "this" context while calling fire function and curcuit breaker uses this parameter while calling the "apply" function.

lance:
I think I would prefer adding a new `apply()` method to to `CircuitBreaker` . For example, something like this.

```js
apply (pThis) {
  // body of current fire() method here
}

fire () {
  this.apply(this.action, arguments);
}
```

PR #374, opened 2019-09-23
feat: add call() method to CircuitBreaker

lance:
This allows callers to control the `this` context when firing
a circuit breaker function.

Fixes: https://github.com/nodeshift/opossum/issues/349

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -151,9 +151,9 @@
                       </a></li>
                       
                       <li><a
-                        href='#circuitbreakerapply'
+                        href='#circuitbreakercall'
                         class='regular pre-open'>
-                        #apply
+                        #call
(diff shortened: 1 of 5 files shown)


lance:
Changed this so that the new circuit method is `call()` instead of `apply()` since semantically it's actually like `Function.prototype.call` instead of `Function.prototype.apply`.

lance:
@lholmquist I used `call()` because it behaves pretty much just like `Function.call()`. And since `Function.call()` is a static function, we aren't overriding anything with this. But I can be persuaded if you can come up with a good name. :)

lholmquist:
I don't have anything in mind,  so call is good :)

helio-frota:
LGTM :+1:

PR #376, opened 2019-10-17
implement `isOurError()`

tjenkinson:
Which provides a way of determining if the rejection from `fire()` or `call()` is from the circuit breaker or the action.

E.g
```js
breaker.fire().catch((e) => {
  if (CircuitBreaker.isOurError(e)) {
    // the system is overloaded
  } else {
    // unexpected error
  }
});

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -18,6 +18,7 @@ const CACHE = new WeakMap();
 const ENABLED = Symbol('Enabled');
 const WARMING_UP = Symbol('warming-up');
 const VOLUME_THRESHOLD = Symbol('volume-threshold');
+const OUR_ERROR = Symbol('our-error');
 const deprecation = `options.maxFailures is deprecated. \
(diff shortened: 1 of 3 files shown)


node-deno-dev:
Hi, what are some of the scenarios where the circuit breaker itself will be in an "error" state? For example: I see logs like "Breaker is open" when isOurError() is true. 

You mention at the top of this page: if (CircuitBreaker.isOurError(e)) {
    **// the system is overloaded** - when exactly does this happen?
  } else {
    // unexpected error
  }

Thanks in advance.

lance:
@node-deno-dev if the circuit is open or shut down, or if function execution times out, or if the function's volume threshold has been exceeded the error is considered to have been generated internally  - in the circuit breaker itself. Otherwise, the error was the function execution throwing an exception. HTH

node-deno-dev:
Thank you Lance for the explanation!

PR #378, opened 2019-10-18
feat: clear intervals on shutdown

scttcper:
after shutting down a circuit breaker, these status intervals are persisted and continue to run. This keeps track of each interval and shuts them down.

Edit: attempting to fix the following warning from jest when using `jest --detectOpenHandles`

```
Jest has detected the following 2 open handles potentially keeping Jest from exiting:

  ●  Timeout



      at new Status (node_modules/opossum/lib/status.js:65:20)
      at new CircuitBreaker (node_modules/opossum/lib/circuit.js:119:20)
```

### DIFF
--- a/lib/status.js
+++ b/lib/status.js
@@ -4,7 +4,7 @@ const WINDOW = Symbol('window');
 const BUCKETS = Symbol('buckets');
 const TIMEOUT = Symbol('timeout');
 const PERCENTILES = Symbol('percentiles');
-const BUCKET_INTERVAL = Symbol('emit-interval');
+const BUCKET_INTERVAL = Symbol('bucket-interval');
 const EMIT_INTERVAL = Symbol('emit-interval');
 
 const EventEmitter = require('events').EventEmitter;
--- a/lib/status.js
+++ b/lib/status.js
@@ -5,7 +5,7 @@ const BUCKETS = Symbol('buckets');
 const TIMEOUT = Symbol('timeout');
 const PERCENTILES = Symbol('percentiles');
 const BUCKET_INTERVAL = Symbol('bucket-interval');
-const EMIT_INTERVAL = Symbol('emit-interval');
+const SNAPSHOT_INTERVAL = Symbol('snapshot-interval');
 
 const EventEmitter = require('events').EventEmitter;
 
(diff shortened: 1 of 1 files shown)


lib/status.js
```diff
@@ -4,6 +4,8 @@ const WINDOW = Symbol('window');
 const BUCKETS = Symbol('buckets');
 const TIMEOUT = Symbol('timeout');
 const PERCENTILES = Symbol('percentiles');
+const BUCKET_INTERVAL = Symbol('emit-interval');
```
tjenkinson:
```suggestion
const BUCKET_INTERVAL = Symbol('bucket-interval');
```
?

scttcper: i think i was deciding between the names emit and bucket. thanks

lib/status.js
```diff
@@ -4,6 +4,8 @@ const WINDOW = Symbol('window');
 const BUCKETS = Symbol('buckets');
 const TIMEOUT = Symbol('timeout');
 const PERCENTILES = Symbol('percentiles');
+const BUCKET_INTERVAL = Symbol('bucket-interval');
+const EMIT_INTERVAL = Symbol('emit-interval');
```
lance: I think this should be `SNAPSHOT_INTERVAL`.
```suggestion
const SNAPSHOT_INTERVAL = Symbol('snapshot-interval');
```

lib/status.js
```diff
@@ -62,20 +64,25 @@ class Status extends EventEmitter {
...
+    }
 
     /**
      * Emitted at each time-slice. Listeners for this
      * event will receive a cumulative snapshot of the current status window.
      * @event Status#snapshot
      * @type {Object}
      */
-    interval = setInterval(_ => this.emit('snapshot', this.stats),
+    this[EMIT_INTERVAL] = setInterval(_ => this.emit('snapshot', this.stats),
```
lance:
```suggestion
    this[SNAPSHOT_INTERVAL] = setInterval(_ => this.emit('snapshot', this.stats),
```

lib/status.js
```diff
@@ -62,20 +64,25 @@ class Status extends EventEmitter {
...
       bucketInterval);
-    if (typeof interval.unref === 'function') interval.unref();
+    if (typeof this[EMIT_INTERVAL].unref === 'function') {
+      this[EMIT_INTERVAL].unref();
```
lance:
```suggestion
    if (typeof this[SNAPSHOT_INTERVAL].unref === 'function') {
      this[SNAPSHOT_INTERVAL].unref();
```

lib/status.js
```diff
@@ -154,6 +161,8 @@ class Status extends EventEmitter {
 
   shutdown () {
     this.removeAllListeners();
+    clearInterval(this[BUCKET_INTERVAL]);
+    clearInterval(this[EMIT_INTERVAL]);
```
lance:
```suggestion
    clearInterval(this[SNAPSHOT_INTERVAL]);
```

scttcper: cheers :+1:

lance:
@scttcper thanks for the contrib!

Issue #371, opened 2019-09-04
IE 11 support / babel


dptoot:
## Node.js Version:
10

## Operating System:
Windows

## Steps to Produce Error:
Use opossum in IE 11.

It looks as though the opossum library isn't running through babel transpilation.  This is causing our IE11 bundles to fail due to many non-supported ES6 features like template-literals and class syntax.

I would think the simple addition of adding a babel transpile, utilizing the babel-preset-env, to the build would suffice for enough support.

Thoughts?

dptoot:
I am able to workaround this today by adding a custom webpack rule to include opossum using our babel-loader.  But I would say that this is usually up to the package to handle es5 support.

tavogel:
This fails for us when using UglifyJS as well. You can see template literals being used in the minified code:
https://cdn.jsdelivr.net/npm/opossum@4.1.0/dist/opossum.min.js

```js
new Error(`Timed out after ${this.options.timeout}ms`)
```

tavogel:
I have opened a PR that works with `babel-loader` and `core-js@3`. Even if you don't want to use the `core-js` polyfills, you can use `babel-loader` and `preset-env` without `useBuiltIns` to at strip the template literals, and other syntactic sugar.

lance:
Fixed in https://github.com/nodeshift/opossum/pull/380

PR #379, opened 2019-10-21
Implement babel-loader in webpack (with polyfill)

tavogel:
Fixes https://github.com/nodeshift/opossum/issues/371

Uses core-js v3 polyfills

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,9 @@
   },
   "homepage": "https://nodeshift.dev/opossum",
   "devDependencies": {
+    "@babel/core": "^7.6.4",
+    "@babel/preset-env": "^7.6.3",
+    "babel-loader": "^8.0.6",
     "codacy-coverage": "^3.4.0",
     "coveralls": "^3.0.3",
     "documentation": "^12.0.0",
@@ -73,5 +76,7 @@
     "hystrix",
     "rate-limiting"
   ],
-  "dependencies": {}
+  "dependencies": {
+    "core-js": "^3.3.3"
+  }
 }
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -58,7 +58,26 @@ function generateConfig (name) {
         }
       })
     ],
-    devtool: 'source-map'
+    devtool: 'source-map',
+    module: {
+      rules: [
+        {
+          test: /\.m?js$/,
+          exclude: /node_modules/,
+          use: {
+            loader: 'babel-loader',
+            options: {
+              presets: [
+                ['@babel/preset-env', {
+                  'useBuiltIns': 'usage',
+                  'corejs': 3
+                }]
+              ]
+            }
+          }
+        }
+      ]
+    }
   };
   return config;
 }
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -73,5 +76,7 @@
     "hystrix",
     "rate-limiting"
   ],
-  "dependencies": {}
+  "dependencies": {
+    "core-js": "^3.3.3"
```
lance: This should be a `devDependency` shouldn't it?

tavogel: Nope! More info:
https://babeljs.io/blog/2019/03/19/7.4.0

The core-js package is a light suite of polyfills for all the major new JS features. With babel `useBuiltIns`, it will dynamically add the imports for any core-js polyfills to the dist folder. These polyfills will bloat the distribution a  tiny bit, but this package will be able to serve a much wider audience as a result. If you prefer to leave the polyfills out, you can remove coreJS and just use babel preset-env, with no args, to transpile only the syntactic sugar, without adding any polyfills. Given that this package is marketed for browsers, I think the overhead is worth it.

Note there's also a second `regenerator-runtime` polyfill dependency that Babel can generate imports for. I didn't include it as a dependency in this PR because you don't use generators or async/await code. If you would prefer maintaining async/await code, then you can easily add the `regenerator-runtime` dependency and a single `require("regenerator-runtime/runtime");` in your entrypoint (like described in the babel blog). This will transpile your bleeding edge ES2019 into browser-native ES5 at build time.

tavogel:
For regression testing, I guess you could do something with Selenium/IE11, BrowserStack, etc. Another possibility would be piping the output through UglifyJS, which only supports ES5 bundles, using the Webpack plugin. This would give you an added benefit of shrinking your dist size. But then you need to test that the Uglified output works too... :)

lholmquist:
Is there something that needs a polyfil specifically?   I think we tried really hard to make this library have no dependencies, so it would be a bummer if we added a dependency to it just for 1 thing

tavogel:
Added an alternative PR with no polyfill: https://github.com/nodeshift/opossum/pull/380

The polyfills installed by babel/core-js are:
```
es.array.concat
es.array.for-each
es.array.from
es.array.is-array
es.array.iterator
es.array.reduce
es.array.slice
es.array.sort
es.date.now
es.date.to-string
es.function.name
es.number.constructor
es.number.is-integer
es.number.max-safe-integer
es.object.create
es.object.define-property
es.object.get-prototype-of
es.object.keys
es.object.set-prototype-of
es.object.to-string
es.promise
es.regexp.exec
es.regexp.to-string
es.string.iterator
es.string.replace
es.symbol
es.symbol.description
es.symbol.iterator
es.weak-map
web.dom-collections.for-each
web.dom-collections.iterator
web.timers
```

It's a huge list because no browserlist was defined. If you put bigger limits on the browsers you target, you get far fewer polyfills.

You can compare the differences between the `/dist/opossum.js` files when you run `npm run build` in the branch for each PR. Quite a big difference!

I agree that since you already have no deps, it should probably stay as lean as possible.

tavogel:
Closed for https://github.com/nodeshift/opossum/pull/380

PR #384, opened 2019-10-31
Set entrypoint to dist/opossum.js

tavogel:
With https://github.com/nodeshift/opossum/pull/380 and https://github.com/nodeshift/opossum/releases/tag/v4.2.0, we are now using Babel to transpile our distribution.

However, I was perplexed... I still get the same error as before, where my app is choking on the use of template literals. I realized the problem is because the distribution folder is not used at all by this library. We ship a transpiled and minified dependency, but don't use either of them by default, because the entrypoint is set to the library path. For the moment I can bypass the error by explicitly using the dist path in my app... `require('opossum/dist/opossum')`.

This pull request updates the package to point to the transpiled distribution of Opossum. Additionally, it only ships the `dist` folder, not the `lib`, so the package size is reduced. I confirmed this fixes my issue as well.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -34,7 +34,7 @@ function generateConfig (name) {
   const config = {
     mode,
     entry: {
-      circuitBreaker: './index.js'
+      circuitBreaker: path.resolve(__dirname, '..', 'lib', 'circuit.js')
     },
     output: {
       path: path.resolve(__dirname, '..', 'dist'),
--- a/index.js
+++ b/index.js
@@ -1,3 +1,0 @@
-'use strict';
-
-module.exports = exports = require('./lib/circuit');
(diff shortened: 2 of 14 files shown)


tavogel:
Any news? My workaround for now is: `const CircuitBreaker = require('opossum/dist/opossum');`

lance:
@tavogel apologies - I have been traveling and this just slipped through the cracks. I'll take a look today.

lance:
@tavogel if you would please update the branch, I'll merge. Thanks.

tavogel:
@lance rebased!

lholmquist:
I had to revert this commit since when i did the 4.2.2 release,  opossum broke for people,  https://github.com/nodeshift/opossum/issues/392,  I think we need to revisit this issue

PR #386, opened 2019-11-05
Fix stuck open

### LINKED ISSUE #385: Simultaneous failures cause circuit to never move into the halfOpened state
## Node.js Version:

10.17.0

## Operating System:

MacOS 10.14.6

## Steps to Produce Error:

I'm seeing an issue that started happening in `4.2.0` (https://github.com/nodeshift/opossum/pull/383) where the circuit never enters the `halfOpened` state.

If you have a circuit that automatically opens after 5 requests and you concurrently launch 6 simultaneous requests that each eventually error:

* The first 4 requests will `fail()` but not call `open()`
* The 5th request will call `open()` which will emit the `open` event and start the timer to later move into the `halfOpened` state
* The 6th request will call `open()` _again_ which will cancel the timer, but since it is already in the `opened` state, will not emit the `open` event and thus not restart the timer.
* The circuit stays in the `opened` state and never enters the `halfOpened` state, so all further requests are rejected.

I think that the correct behavior might be for `open()` to cancel the current timer and immediately re-add it, so that it pushes the timer end date out regardless of whether or not the state change happens?

tjenkinson:
fixes #385 by changing the behaviour of `open()` back to being a noop if the state is already open


d-tylczynski-foodpanda:
We just wanted to report he issue, thanks for fixing it

Acanguven:
Thanks for fixing, I am also waiting for this PR to be merged.

Issue #388, opened 2019-11-25
Circuit breaker gets stuck in halfOpen state when using errorFilter


jairelton:
## Node.js Version:
10.15.0
## Operating System:
MacOS 10.15.1
## Steps to Produce Error:
I'm experiencing an issue when trying to use `errorFilter` where the circuit gets stuck in the `halfOpen` state.

If you have a circuit in the `halfOpen` state and the subsequent failure is filtered by the error filter, the circuit remains in the `halfOpen` state, with `pendingClose === false` and no pending requests. This way it'll never go back to `closed`.

Step by step:

- The circuit reaches the error threshold and calls `open()`
- After the reset timeout, the circuit breaker goes into `halfOpen` state and `PENDING_CLOSE = true`
- The subsequent request changes `PENDING_CLOSE` to `false`
- The request fails and calls `failure()`
- The errorFilter filters the error and the failure is ignored
- The circuit breaker is now in an inconsistent state

I just sent a PR with a possible solution for this issue where I ignore the errorFilter when in HALF_OPEN state.

lholmquist:
closing.  fixed in #389   released as 4.2.2

PR #389, opened 2019-11-25
Fix stuck halfOpen when using errorFilter

jairelton:
Fix issue #388

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -643,10 +643,7 @@ function handleError (error, circuit, timeout, args, latency, resolve, reject) {
 
   if (circuit.options.errorFilter(error)) {
     circuit.emit('success', error, latency);
-    return reject(error);
-  }
-
-  fail(circuit, error, args, latency);
+  } else fail(circuit, error, args, latency);
 
   const fb = fallback(circuit, error, args);
   if (fb) resolve(fb);
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -663,7 +663,7 @@ function fallback (circuit, err, args) {
 }
 
 function fail (circuit, err, args, latency) {
-  if (circuit.options.errorFilter(err)) return;
+  if (!circuit.halfOpen && circuit.options.errorFilter(err)) return;
 
   /**
    * Emitted when the circuit breaker action fails
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -640,7 +640,11 @@ class CircuitBreaker extends EventEmitter {
 
 function handleError (error, circuit, timeout, args, latency, resolve, reject) {
   clearTimeout(timeout);
-  fail(circuit, error, args, latency);
+
+  if (circuit.options.errorFilter(error)) {
+    circuit.emit('success', error, latency);
+  } else fail(circuit, error, args, latency);
+
   const fb = fallback(circuit, error, args);
   if (fb) resolve(fb);
   else reject(error);
@@ -663,8 +667,6 @@ function fallback (circuit, err, args) {
 }
 
 function fail (circuit, err, args, latency) {
-  if (!circuit.halfOpen && circuit.options.errorFilter(err)) return;
-
   /**
    * Emitted when the circuit breaker action fails
    * @event CircuitBreaker#failure
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -643,7 +643,10 @@ function handleError (error, circuit, timeout, args, latency, resolve, reject) {
 
   if (circuit.options.errorFilter(error)) {
     circuit.emit('success', error, latency);
-  } else fail(circuit, error, args, latency);
+    return reject(error);
+  }
+
+  fail(circuit, error, args, latency);
 
   const fb = fallback(circuit, error, args);
   if (fb) resolve(fb);
--- a/test/half-open-test.js
+++ b/test/half-open-test.js
@@ -2,7 +2,7 @@
 
 const test = require('tape');
 const CircuitBreaker = require('../');
-const { timedFailingFunction } = require('./common');
+const { timedFailingFunction, failWithCode } = require('./common');
 
 test('When half-open, the circuit only allows one request through', t => {
   t.plan(10);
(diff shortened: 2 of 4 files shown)


lance:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/half-open-test.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/opossum/opossum/pullRequest?prid=4563987&bid=15344776)

lib/circuit.js
```diff
@@ -663,7 +663,7 @@ function fallback (circuit, err, args) {
 }
 
 function fail (circuit, err, args, latency) {
-  if (circuit.options.errorFilter(err)) return;
+  if (!circuit.halfOpen && circuit.options.errorFilter(err)) return;
```
lance: I meant to include this in my review. This seems wrong because, for example, in your test, the filter will ignore anything with an error code < 500 - except in the case where the circuit is in the half-open state. I think it would be better to reset `PENDING_CLOSE`.
```suggestion
  if (!circuit.halfOpen && circuit.options.errorFilter(err)) {
    if (circuit.halfOpen) circuit[PENDING_CLOSE] = true;
    return;
  }
```

jairelton: That was my first implementation, but it seemed good neither because the circuit could accept a lot of requests while in the half-open state. But thinking again it seems a better solution indeed.

Another option would be going straight to the closed state, as a filtered error usually is an expected result. What do you think?

lance: > Another option would be going straight to the closed state, as a filtered error usually is an expected result. What do you think?

Actually, this is a good point. If there is an error filter in place, when an error is filtered it is considered a success condition. The circuit should probably be opened at that point.

jairelton: If the error should be considered a success, shouldn't we emit `success` instead of `failure`?

```suggestion
  if (circuit.options.errorFilter(err)) {
    circuit.emit('success', err, latency);
    return;
  }
```
But err as result doesn't seem right, does it?

lance: I have to think about this. You are right, `success` should probably be emitted - though that has never been the case with filtered errors in the past. Not sure whether passing the `Error` object makes sense.

lance: I think the circuit should emit `success` and provide the error. It's a little weird, but providing the error at least gives some information.

test/half-open-test.js
```diff
@@ -46,3 +46,38 @@ test('When half-open, the circuit only allows one request through', t => {
...
+    t.ok(breaker.pendingClose, 'should be pending close after timeout');
+    breaker
+      .fire(400) // fail with a filtered error
+      .catch(e =>
+        t.equals(e.message, 'Failed with 400 status code', 'should fail again'))
+      .then(() => {
+        t.ok(breaker.opened,
+          'should be opened after failing with filtered error');
+      })
+      .then(_ => breaker.shutdown())
```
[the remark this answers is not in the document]
jairelton: The problem is that I would need to bind it.

`.then(breaker.shutdown.bind(breaker))`

lib/circuit.js
```diff
@@ -640,7 +640,11 @@ class CircuitBreaker extends EventEmitter {
 
 function handleError (error, circuit, timeout, args, latency, resolve, reject) {
   clearTimeout(timeout);
-  fail(circuit, error, args, latency);
+
+  if (circuit.options.errorFilter(error)) {
+    circuit.emit('success', error, latency);
+  } else fail(circuit, error, args, latency);
+
   const fb = fallback(circuit, error, args);
```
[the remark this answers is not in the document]
jairelton: I included a new test to check this.

lance:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/error-filter-test.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/nodeshift/opossum/pullRequest?prid=4563987&bid=15344776)

lance:
@jairelton thanks for making the change. In reviewing it, I realized that past behavior has been that, whether or not the error is filtered, `fallback()` is called. And if that call succeeds, then a resolved promise is returned. If not, a rejected promise is returned.

What I have suggested you do is changing the the behavior, and now I am having second thoughts about it because this would really be a `semver-major` change, I think. @lholmquist do you have any thoughts on this?

lance:
@lholmquist do you mind taking a look and letting me know what you think?

lib/circuit.js
```diff
@@ -640,7 +640,14 @@ class CircuitBreaker extends EventEmitter {
 
 function handleError (error, circuit, timeout, args, latency, resolve, reject) {
   clearTimeout(timeout);
+
+  if (circuit.options.errorFilter(error)) {
+    circuit.emit('success', error, latency);
+    return reject(error);
```
lholmquist: if you've filtered the error,  and emitting success,  why would you reject the promise.  i thought this would have resolved?

lance: That is a good question. It's complicated.

In the current version, if the error is filtered, but there is no fallback function, no `failure` event is emitted, but a rejected `Promise` is returned. If there is a fallback function, a resolved `Promise` is always returned. In this case, the behavior has been changed so that the fallback function is never called in the event of a filtered error. To remain partially consistent with past behavior, then the promise should be rejected since no fallback is applied.

lholmquist: it just seems weird to emit success but reject

lance: Agree... what do you think the correct approach is? Always resolve when an error is filtered? I'm OK with that, but it will be at least a minor version bump.

jairelton: @lance Do you mean resolve when there's a fallback?

lholmquist: I would vote for, if a error happens, but there is an errorFilter setup, then we should be resolving

If it has to be a major bump, thats ok.  Is error filter a fairly new addition.  i can't remember when it was added

jairelton: @lholmquist But resolve with what result? If there's a fallback we resolve with the fallback result, but if there isn't, then there is no result to resolve with, in this case, I think we should reject with the error, so the application can handle it.

lholmquist: Yes, i think that sounds correct

lance: I think I'm good with this (until I'm not - lol).

lance:
@lholmquist I'm going to land this for now, but let's keep an eye on how filters/fallbacks are used going forward.

PR #402, opened 2020-03-19
chore: add deprecation warnings and errors flags

### LINKED ISSUE #322: Add deprecations warnings and failures to CI builds
Node provides flags that allow for failing when deprecated functionality is used. There are also flags for printing out deprecation warnings on functionality that will soon be deprecated. Turning on these flags will help us catch issues early that may cause failures in the future.

lholmquist:
fixes #322

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,12 @@ build-common: &common-build
         key: v1-dependencies-{{ checksum "package.json" }}
 
     # run tests!
-    - run: npm run ci
+    - run:
+        name: Run Tests
+        command: npm run ci
+        # environment:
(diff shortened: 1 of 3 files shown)


lholmquist:
this is a work in progress since circle ci should fail with the given env variables

lholmquist:
The failure is because we are creating docs as part of the `npm run ci` call and the docs are using a deprecated API.  

Since this isn't related to Opossum itself, i think we can not do the doc build in the ci tests,  which should get around those issues

lholmquist:
The reason why node 12 is failing here is becuase of this issue, https://github.com/istanbuljs/spawn-wrap/issues/89

Not really anything we can do about it until is fixed

PR #424, opened 2020-05-21
Removing redundant entries

alexalikiotis:


alexalikiotis:
> What is this `opossum-5.0.0.tgz` file that made it in?

Ooops, that was me testing the **npm pack** command that helio said but probably forgot to remove it before the commit 🤦‍♂️

PR #443, opened 2020-07-16
fix(test): Avoiding an UnhandledPromiseRejection

helio-frota:
This commit avoids an UnhandledPromiseRejectionWarning to be throwing
from `'Breaker resets after a configurable amount of time when multiple
jobs fail'` test.


helio-frota:
@lholmquist thanks for the review

PR #454, opened 2020-08-18
doc: fixed the doc issue about allowWarmUp and rollingCountTimeout

### LINKED ISSUE #453: Missing/incorrect info in the doc
Hello there!

Just noticed that `rollingCountDuration` isn't present in the official documentation [there](https://nodeshift.dev/opossum/#circuitbreaker).
As for `allowWarmUp`, here's the description:
> determines whether to allow failures without opening the circuit during a brief warmup period (this is the rollingCountDuration property). Default: false allow before enabling the circuit. This can help in situations where no matter what your errorThresholdPercentage is, if the first execution times out or fails, the circuit immediately opens. Default: 0 

As it has a boolean value, I find it weird that the default is `0` and that there's another default `false` in the same text.

MathieuAA:
Fixes #453

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -64,10 +64,10 @@ Please use options.errorThresholdPercentage`;
  * construction. Default: true
  * @param {boolean} options.allowWarmUp determines whether to allow failures
  * without opening the circuit during a brief warmup period (this is the
- * `rollingCountDuration` property). Default: false
+ * `rollingCountTimeout` property). Default: false
  * allow before enabling the circuit. This can help in situations where no
  * matter what your `errorThresholdPercentage` is, if the first execution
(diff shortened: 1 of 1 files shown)


lholmquist:
The only file that should be updated here is the `lib/circuit.js` file. When we publish a new version,  the docs will also get updated.  There are also some rogue code formatting that shouldn't be included

helio-frota:
@lholmquist that's correct thanks :+1:  

@MathieuAA can you please update your PR ? Thanks

MathieuAA:
Will do, thanks for the review.

MathieuAA:
Done.

lholmquist:
Thanks @MathieuAA !!

Issue #459, opened 2020-08-26
Feature Request: fallback function should pass the same parameters as fire


tomkoufakis:
FEATURE REQUEST:

When configuring a fallback function, the method signature is without parameters. For example:

circuit.fallback(**()** => { ... });

The fallback function should be able to accept the same parameters as the fire function. The fallback function, when called, should be called with these same parameters.

This will allow for parameters that can mutate the response of the fallback function. 

Example usage:

// circuit setup, only called once
circuit.fallback((arg1, arg2, arg3) => {
  return arg1 + arg2 + arg3;
});

// called for each request
return circuit.fire(arg1, arg2, arg3).then((response) => {
  console.log(JSON.stringify(response));
});

At present, I do not have a good workaround for this.

tomkoufakis:
After looking through the opossum source, it appears this is already supported and just the _documentation needs to be updated_.

helio-frota:
@tomkoufakis feel free to send a PR.

Either way I created a example usage here:
https://github.com/nodeshift-starters/opossum-playground/tree/master/debuglog/core/fallback-parameters

When running `npm start` it shows:

```
DBG-CORE 17491: inside the constructor
DBG-CORE 17491: sets timeout to 10000
DBG-CORE 17491: sets resetTimeout to 30000
DBG-CORE 17491: creates new semaphore with capacity 9007199254740991
DBG-STATUS 17491: status created
DBG-CORE 17491: called fire function with args 20000,1,2,3   <---------
DBG-CORE 17491: fire event emitted
DBG-SEMAPHORE 17491: semaphore counter 9007199254740991
DBG-SEMAPHORE 17491: semaphore released
DBG-CORE 17491: timeout event emitted
DBG-CORE 17491: failure event emitted
DBG-CORE 17491: open event emitted
Sorry, out of service right now. But your parameters are: 20000, 1, 2 and 3    <---------
DBG-CORE 17491: fallback event emitted
fires 1 fallbacks 1
```

PR #460, opened 2020-09-10
Adding docs about fallback parameters

helio-frota:
Related to #459 

@lance @lholmquist I'm not 100% sure if we need to run `npm run build:docs` before to send the PR when updating the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,21 @@ the `open` state.
 When a fallback function is triggered, it's considered a failure, and the
 fallback function will continue to be executed until the breaker is closed.
 
+The fallback function accepts the same parameters as the fire function:
+
+```javascript
+const delay = (delay, a, b, c) =>
+  new Promise((resolve) => {
+    setTimeout(() => {
+      resolve();
+    }, delay);
+  });
+
+const breaker = new CircuitBreaker(delay);
+breaker.fire(20000, 1, 2, 3);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,21 @@ the `open` state.
 When a fallback function is triggered, it's considered a failure, and the
 fallback function will continue to be executed until the breaker is closed.
 
+Also the fallback function accepts the same parameters as the fire function:
```
lance, line 79: Nit:
```suggestion
The fallback function accepts the same parameters as the fire function:
```

lance (approved):
@helio-frota - yes, if you are updating docs, `npm run build` is needed. The site is published from the `master` branch in the `/docs` directory.

lholmquist (approved):
LGTM

Regarding having to generate the docs,  i wonder if we should start using a GH action to generate them on merges to master?

README.md
helio-frota: Thank you. PR updated :+1:

lholmquist: I believe when we do a release,  the docs are updated as part of that npm script,  so that should take care of it.  I've created an issue for trying to automate the doc creation also,  #461 

We could also see about using release-please for doing releases,  similar to what we are about to do on the js cloudevents SDK library, https://github.com/cloudevents/sdk-javascript/pull/329

Issue #489, opened 2020-11-11
Additional param in errorFilter option


vizhukova:
Hello, opossum team!

I've got a question about expansion `errorFilter` functionality.
**Initial data:** 
The breaker once defined in the project and reused by every request
I need to ignore errors only for one request (as it is expected behavior and I need to call query until it response success)
So I expect something like that
```
const breaker = new CircuitBreaker(asyncFunction, {
      name,
      timeout,
      resetTimeout,
      errorFilter: (err, query) => {
       if(err.code === 'SOME_ERROR_CODE' && query === 'some')
     },
    })
```

Is it possible to add another one `query` param to the `errorFilter`? 

Thanks in advance,
Veronika

lance:
Hi @vizhukova it's not clear what you expect `query` to be in this case. Would it work to pass the parameters originally provided to the failing `asyncFunction`? Specifically here in `handleError`:

https://github.com/nodeshift/opossum/blob/32762f4162dbe1a1feffa7ef97c3e2f86cec977c/lib/circuit.js#L648-L655

If line 651 looked instead like

```js
  if (circuit.options.errorFilter(error, args)) {
```
Would this solve the problem?

vizhukova:
Yeah! This will help

PR #492, opened 2020-11-17
feat: pass circuit params to error filter

lance:
When the circuit invocation fails, and there is an error filter in place,
this change provides the function invocation parameters as the second
parameter to `errorFilter()`.

Fixes: https://github.com/nodeshift/opossum/issues/489


alexalikiotis (approved):
LGTM

helio-frota (approved):

PR #496, opened 2020-11-18
Update README.md

elderbas:
I believe this should be `breaker` instead?


lance (approved):
lgtm - thanks for the catch!

helio-frota (approved):

PR #499, opened 2020-11-23
fix: lint issues in test

helio-frota:

PR #502, opened 2020-11-24
CI fixes

richardlau:
- Revert 814cd4f4f10796a475ff7b032a44a822fef0bdae as this makes the CI job run with the base commit of the master branch and not the pull request head (see https://github.com/nodeshift/opossum/pull/499#issuecomment-732126270).
- Switch to using the coveralls GitHub action. Will need to do a follow up test to see if pull requests from forks behave better than before 814cd4f4f10796a475ff7b032a44a822fef0bdae but we should be no worse off.

### DIFF
--- a/.github/workflows/nodejs-ci-action.yml
+++ b/.github/workflows/nodejs-ci-action.yml
@@ -24,5 +24,20 @@ jobs:
     - run: npm ci
     - run: npm test
     - run: npx @pkgjs/support validate
-    - run: echo 'repo_token:' ${{ secrets.COVERALLS_GITHUB_TOKEN }} > ./.coveralls.yml
-    - run: node_modules/nyc/bin/nyc.js report --reporter=text-lcov | node_modules/coveralls/bin/coveralls.js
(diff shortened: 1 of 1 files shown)


richardlau:
@helio-frota I'd like to land this PR and then rebase the branch for https://github.com/nodeshift/opossum/pull/499 onto the updated master (which should 🤞 give us a green CI on that PR).

helio-frota:
@richardlau :+1:

richardlau:
> * Switch to using the coveralls GitHub action. Will need to do a follow up test to see if pull requests from forks behave better than before [814cd4f](https://github.com/nodeshift/opossum/commit/814cd4f4f10796a475ff7b032a44a822fef0bdae) but we should be no worse off.

I opened a test pull request from my fork of opossum and the checks passed, which is encouraging: https://github.com/nodeshift/opossum/pull/503/checks
I can't say for definite how it'll behave for someone who has forked this repository that doesn't have write access to it but we can deal with that as/when it happens if any issues arise.

Issue #509, opened 2020-12-04
Integrating opossum with axios


matteodisabatino:
Hi,
I'm trying to integrate oppossum with axios but I noticed a strage behaviour (in my opinion):

I'm trying to make an HTTP request that **must** fail with and without opossum.

Without opossum I make the request this way and it works as excepted:

```javascript
const main = async () => {
  try {
    const response = await axios({
      method: 'POST',
      url: 'http://www.mywebsite.com/'
    })

    console.log('SUCCESS')
    console.log('response: ', response)
  } catch (e) {
    console.log('FAILURE')
    console.log('e: ', e) // code goes here and that's correct
  }
}

main()
  .catch(ex => {
    console.log('Uncaught excpetion: ', ex)
    process.exit(1)
  })
```

Instead _with_ opossum I make the request this way:

```javascript
const makeHTTPRequest = args => {
  const breaker = new CircuitBreaker(axios, {
    timeout: 3000,
    errorThresholdPercentage: 50,
    resetTimeout: 30000
  })

  breaker.fallback(fArgs => console.log('FUNCTIONC FALLBACK, fArgs: ', fArgs))

  breaker.on('halfOpen', fArgs => console.log('EVENT halfOpen, fArgs: ', fArgs))
  breaker.on('close', fArgs => console.log('EVENT close, fArgs: ', fArgs))
  breaker.on('open', fArgs => console.log('EVENT open, fArgs: ', fArgs))
  breaker.on('shutdown', fArgs => console.log('EVENT shutdown, fArgs: ', fArgs))
  breaker.on('fire', fArgs => console.log('EVENT fire, fArgs: ', fArgs))
  breaker.on('cacheHit', fArgs => console.log('EVENT cacheHit, fArgs: ', fArgs))
  breaker.on('cacheMiss', fArgs => console.log('EVENT cacheMiss, fArgs: ', fArgs))
  breaker.on('reject', fArgs => console.log('EVENT reject, fArgs: ', fArgs))
  breaker.on('timeout', fArgs => console.log('EVENT timeout, fArgs: ', fArgs))
  breaker.on('success', fArgs => console.log('EVENT success, fArgs: ', fArgs))
  breaker.on('semaphoreLocked', fArgs => console.log('EVENT semaphoreLocked, fArgs: ', fArgs))
  breaker.on('healthCheckFailed', fArgs => console.log('EVENT healthCheckFailed, fArgs: ', fArgs))
[...]

PR #510, opened 2020-12-04
fix: catch exceptions in fallback functions

lance:
When a fallback function is called and throws an exception, the circuit
breaker should catch the exception and return a rejected Promise.

See: https://github.com/nodeshift/opossum/issues/509


helio-frota (approved):
LGTM

lholmquist (approved):
LGTM,  @lance thoughts on needing a major release for this?  I don't think so,  but i can see an argument for needing it.  although,  it is just fixing what should have been there

lance:
I think it's really more of a bug fix...

Issue #514, opened 2020-12-15
Default value for action in constructor


JimFung:
https://github.com/nodeshift/opossum/blob/208495798b25c1194d95157c7a02db6dbbf649a3/lib/circuit.js#L138

Pretty unlikely scenario but if `action` is undefined and no name is provided for options in the constructor, then its possible to hit the following error:
```
TypeError: Cannot read property 'name' of undefined
```

As for a proposed fix, I'm unsure what a reasonable fallback value would be for action or if another solution (i.e. the user must pass a valid value for action) would be appropriate, I defer to the maintainers' expertise.

helio-frota:
@JimFung thanks for catching that. PR sent with a fix :+1:

PR #522, opened 2020-12-21
fix: using default parameter to avoid runtime error

helio-frota:
This is a quick fix to avoid the following error when no action is passed in the constructor:

TypeError: Cannot read property 'name' of undefined

Related to #514

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -110,7 +110,7 @@ class CircuitBreaker extends EventEmitter {
     return !!error[OUR_ERROR];
   }
 
-  constructor (action, options = {}) {
+  constructor (action = {}, options = {}) {
     super();
     this.options = options;
     this.options.timeout =


mhdawson (approved):
LGTM

lance (commented):
The `action` parameter should be a function not `object`. However, personally I don't think a default action should exist as it makes no sense. In my opinion, a `TypeError` should still be thrown, but with an error message describing the problem. For example, "No action provided. Cannot construct a CircuitBreaker without an invokable action."

helio-frota:
> I don't think a default action should exist as it makes no sense. In my opinion, a TypeError should still be thrown, but with an error message describing the problem. For example, "No action provided. Cannot construct a CircuitBreaker without an invokable action."

makes a lot of sense to me looking this way now. I'm going to create an issue for this.

PR #563, opened 2021-04-08
docs: Fix documentation for default value of rollingPercentilesEnabled

stephenkubovic:
This PR is a very small documentation fix. The `rollingPercentilesEnabled` option is documented as defaulting to false, but rolling percentiles are [enabled by default](https://github.com/nodeshift/opossum/blob/16341bb806c14fec29cb25e7bd301e975ae23631/lib/circuit.js#L123) (and the corresponding [test](https://github.com/nodeshift/opossum/blob/16341bb806c14fec29cb25e7bd301e975ae23631/test/test.js#L811)) unless `rollingPercentilesEnabled` is explicitly set to `false`.


lholmquist (approved):
Nice catch!

lance (approved):

Issue #564, opened 2021-04-12
Current behavior of the errorFilter option


mNalon:
## Node.js Version:
v12.21.0

## Operating System:
macOS

## Steps to Produce Error:
In the last release, when I have an action that returns a rejected promise with an error that I want to filter, the output of the fire method becomes a resolved promise with the truthy value returned by the errorFilter callback option. Is this intentional?

I believe this is happening because of the change added by this line https://github.com/nodeshift/opossum/blob/7a920bfada0cc21a8c87c2e453da91da9b9fe242/lib/circuit.js#L661

If this change was intentional, how can I get the error thrown by the action in the case of an error that is being filtered by the errorFilter?

lholmquist:
this change was made intentionally.  The thought was that if there is an error, but you have a error filter that passes, then that really isn't an error and the the fallback function shouldn't be called, which was happening before the change was made.

The error should still be thrown if the error filter fails,  if that doesn't happen, then there is a bug somewhere

mNalon:
Hi, @lholmquist, thank you for replying quickly!

In my case, I don't have a fallback function. I used to use the filterError only to ignore some errors and don't consider them to open the circuit. But still, I depend on these errors being exposed by the fire method to take the correct actions out of the circuit breaker logic. 

Do you have any suggestions on how I  could keep this behavior?

lance:
> But still, I depend on these errors being exposed by the fire method to take the correct actions out of the circuit breaker logic.

I'm not clear on what you mean by "these errors being exposed by the fire method". You do still have access to the errors themselves in your `errorFilter()` function. Would it help for the resolved promise to contain both a truthy value, and the caught exceptions? E.g.

```js
function handleError (error, circuit, timeout, args, latency, resolve, reject) {
  let errorFiltered;
  if ((errorFiltered = circuit.options.errorFilter(error, ...args))) {
    circuit.emit('success', error, latency);
    resolve({ err: error, filtered: errorFiltered);
  }
  // ... etc
}
```

mNalon:
Hi, @lance, thank you for your reply!

What I meant was related to the fire function. 

In the previous release, this code:

```
cb = new CircuitBreaker(
  () => Promise.reject('error that should not increment the failure statistic'), 
  { errorFilter: (err) => true } // filtering everything
)

cb.fire().then(console.log).catch(console.error)
```

[...]

PR #567, opened 2021-04-15
fix: return error from invocation when error filter filtered errors

lance:
This commit reverts some of the behavior changes in
https://github.com/nodeshift/opossum/pull/556. In that PR, the intent
was to avoid calling the fallback function when an an invocation error
is filtered by a user supplied `errorFilter` function. It did
accomplish that, however, it also changed how a function resolves
in the case of a filtered error. Previously, even though the error
was filtered, the function invocation would return the error to the
caller. With #556, this changed so that the caller instead receives
simply a truthy value.

This commit reverts that behavior so that the error is returned.

While it may seem counter intuitive to return an error when the it
was filtered by a user supplied `errorFilter` function, there are good
reasons to do so. It provides the caller with error information at the
point of invocation instead of in an event listener which may be
disconnected from the invocation code path.

The purpose of `errorFilter` is simply to prevent filtered errors from
causing the circuit to open. But the fact is that the function invocation
failed, and providing this to the user at the point of failure is better
usability in my opinion.

Plus, it's what we've always been doing, and I think the change to returning
a truthy value was really just a side effect of not calling the fallback
function. My preference would be to minimize the breaking changes in 6.x,
and this PR helps to accomplish that (albeit 6.0 will be a weird bump in
the road).

Fixes: https://github.com/nodeshift/opossum/issues/564

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -655,18 +655,20 @@ class CircuitBreaker extends EventEmitter {
 function handleError (error, circuit, timeout, args, latency, resolve, reject) {
   clearTimeout(timeout);
 
-  let errorFiltered;
-  if ((errorFiltered = circuit.options.errorFilter(error, ...args))) {
+  if (circuit.options.errorFilter(error, ...args)) {
+    // The error was filtered, so emit 'success'
     circuit.emit('success', error, latency);
-    resolve(errorFiltered);
   } else {
+    // Error was not filtered, so emit 'failure'
     fail(circuit, error, args, latency);
 
-    // Only call the fallback function if there errorFilter doesn't succeed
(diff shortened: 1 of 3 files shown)


lance:
Note that even though this is a breaking change from 6.0 behavior, I think the change introduced in #564 should not have had the effect it did on circuit return values (and I know, I reviewed and signed off on that PR). Therefore, I don't think this should be a major version bump. In fact, I think it should probably be patch level.

lholmquist (approved):
That explanation does make sense.   And yea,  this can just be a patch or minor release

lance:
I cannot figure out the best title for this PR! It's like a tongue twister with recursion. An Escher-like PR title.

Issue #504, opened 2020-12-01
Question - Serverless Integration with a generic stateful endpoint.  


jamesvillarrubia:
Really interested in this library, but, as I understand it, it maintains state as property within the Class object.  This is fine server-side, if I can share that Class object across multiple requests, but becomes a multiplicative problem in a high-concurrency, stateless environment like AWS Lambda.  

1) Is there a pattern you recommend for serverless environments or sharing object state across containers that I'm not seeing in the docs?

2) If not, would that be a plugin/addition/alternative that the nodeshift team would be interested in?

lance:
@jamesvillarrubia I've also been thinking about this module in a stateless environment. As you mention, in an environment where instances are constantly being spun up/down, state doesn't really make much sense. It almost seems as though there should be some kind of proxy service that all of the stateless functions are pointing at. In this proxy service is where the circuit breaker lives. Of course, that doesn't help when the proxy is down. Functions, being inherently stateless, just can't maintain any information about the status of the remote endpoint - or at least not enough to be of much use.

Not sure this helps much...

> If not, would that be a plugin/addition/alternative that the nodeshift team would be interested in?

I can't speak for the team any more, but if you want to elaborate a bit on what you are thinking that would help.

mlevkovsky:
@lance hopping on this thread. 
It would be cool if we could have a feature where we can configure a remote store for the state of the CB. For example, https://github.com/gunnargrosch/circuitbreaker-lambda, allows the user to configure dynamodb to maintain the state of the circuit breaker.
Ideally, if you pass in some sort of flag, (a connection string and key for example) then this package would know to go retrieve it's information from a remote source.

jamesvillarrubia:
@mlevkovsky Yes that's exactly what I was imagining.  We just ported that same library you linked to firestore in GCP, but a generic storage API (with some adapters) would make me switch back to opossum, with all nice bells and whistles that it has.

lholmquist:
@mlevkovsky interesting concept.  I would want to try and keep the main Opossum module as stateless as possible,  so perhaps this is something that either plugins into opossum or that opossum can send its info too.  We've done something similar with our opossum-promethesus module, https://github.com/nodeshift/opossum-prometheus .

mlevkovsky:
@lholmquist that makes sense. Thanks for linking me to that repo! much appreciated.

@jamesvillarrubia  Further more, I actually made an interesting discovery about lambda, that lets opossum works right out of the box.

[...]

PR #568, opened 2021-04-19
feat: Add the ability to prime a breaker with previous stats

lholmquist:
For this first pass, i've only added the ability to import a previous stats object in.  

I'm not sure if we need to import the state of a breaker(closed, open, half-open, etc...) but that is something we can iterate on.

relates to #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ breaker.fallback((delay, a, b, c) => `Sorry, out of service right now. But your 
 ```
 ### Status Initialization
 
-There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  The major use case for this is in a serverless functions environment,  or in a container based deployment, where the container being deployed is ephemeral.
+There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  Primary use cases for this are in a serverless environment such as Knative or AWS Lambda, or any container based platform, where the container being deployed is ephemeral.
 
 Getting the existing cumalative stats for a breaker can be done like this:
 
--- a/README.md
+++ b/README.md
@@ -87,6 +87,56 @@ const breaker = new CircuitBreaker(delay);
 breaker.fire(20000, 1, 2, 3);
 breaker.fallback((delay, a, b, c) => `Sorry, out of service right now. But your parameters are: ${delay}, ${a}, ${b} and ${c}`);
 ```
+### Status Initialization
+
+There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  The major use case for this is in a serverless functions environment,  or in a container based deployment, where the container being deployed is ephemeral.
+
+Getting the existing cumalative stats for a breaker can be done like this:
+
+```
+const stats = breaker.status.stats;
+```
+
+`stats` will be an object that might look similar to this:
+
+```
+{
+  failures: 11,
+  fallbacks: 0,
+  successes: 5,
+  rejects: 0,
+  fires: 16,
+  timeouts: 0,
+  cacheHits: 0,
+  cacheMisses: 0,
+  semaphoreRejections: 0,
+  percentiles: {
+    '0': 0,
+    '1': 0,
+    '0.25': 0,
+    '0.5': 0,
+    '0.75': 0,
+    '0.9': 0,
+    '0.95': 0,
+    '0.99': 0,
+    '0.995': 0
+  },
+  latencyTimes: [ 0 ],
+  latencyMean: 0
+}
+```
+
+To then re-import those stats, first create a new `Status` object with the previous stats and then pass that as an option to the CircuitBreaker constructor:
+
+```
+const statusOptions = {
+  stats: {....}
+};
+
+const newStatus = CircuitBreaker.newStatus(statusOptions);
+
+const breaker = new CircuitBreaker({status: newStatus});
+```
 
 ### Browser
 
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ There may be times where you will need to pre-populate the stats of the Circuit 
 Getting the existing cumalative stats for a breaker can be done like this:
 
 ```
-const stats = breaker.status.stats;
+const stats = breaker.stats;
 ```
 
 `stats` will be an object that might look similar to this:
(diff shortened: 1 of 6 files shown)


lance:
Looking good!

> I'm not sure if we need to import the state of a breaker(closed, open, half-open, etc...) but that is something we can iterate on.

If we do this, things get a bit messier. But I think it's probably the right thing to do. Imagine where you're app is under great load, and more and more instances are being scaled up automatically. Now imagine that a circuit in the app has been consistently failing. Each time a new node spins up it will hit the broken endpoint at least once, partially undoing the benefits of the circuit breaker. Obviously, however, the timeout period for entering `HALF_OPEN` would be reset anew with each node that spins up, because I don't think this value can be persisted in a way that lets new instances wait for only as long as the timeout period minus however much time has already elapsed.

I wonder, if instead of `circuit.importStats()` we make it a little more generic. E.g. `circuit.initalize(stats, state)`?

lholmquist:
> I wonder, if instead of circuit.importStats() we make it a little more generic. E.g. circuit.initalize(stats, state

I'm good with that.

And I agree that we should probably be initializing breakers to the state in which they where when the function/pod/etc..  was destroyed.  I might make that a separate PR so this one can lay the groundwork and not get to bloated

lholmquist:
I update the `initialize` function signature to be an object,  so it will be easier to pass in both the stats and the state in the future

lance:
> I update the `initialize` function signature to be an object, so it will be easier to pass in both the stats and the state in the future

Why not just accept a `Status` object with `status` as an option in the constructor?

```js
const Circuit = require('opossum')

// Because we only export the CircuitBreaker class we have to do this kind of clunky stuff
const status = Circuit.newStatus({ stats })
const circuit = new Circuit(myfn, { status })
```

This would allow us to do some sanity checks when the Status instance is initalized. What do you think? Too clunky?

lholmquist:
That first example doesn't look to clunky and i would like to not have to do a Major bump if we don't have to

Let me refactor what i did to see how it looks

I would rather save that for when we convert to a pure ESM module :)

helio-frota (approved):
LGTM.

lholmquist:
@lance just pushed a refactor.

now it looks like this:

```
const statusOptions = {}
statusOptions.stats = {.....}
const newStatus = CircuitBreaker.newStatus(statusOptions);

const breaker = new CircuitBreaker(fn, {status: newStatus})

```


I still need to update the readme,  but how is this looking.  I like this better than having that initialization function.

lholmquist:
@lance @helio-frota  i think i'm ready for a review here with this new refactor.

README.md
```diff
@@ -87,6 +87,56 @@ const breaker = new CircuitBreaker(delay);
...
 breaker.fallback((delay, a, b, c) => `Sorry, out of service right now. But your parameters are: ${delay}, ${a}, ${b} and ${c}`);
 ```
+### Status Initialization
+
+There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  The major use case for this is in a serverless functions environment,  or in a container based deployment, where the container being deployed is ephemeral.
+
+Getting the existing cumalative stats for a breaker can be done like this:
+
+```
+const stats = breaker.status.stats;
```
lance, line 97:
```suggestion
const stats = breaker.stats;
```

There's actually a convenience function for this on `Circuit`

README.md
```diff
@@ -87,6 +87,56 @@ const breaker = new CircuitBreaker(delay);
 breaker.fire(20000, 1, 2, 3);
 breaker.fallback((delay, a, b, c) => `Sorry, out of service right now. But your parameters are: ${delay}, ${a}, ${b} and ${c}`);
 ```
+### Status Initialization
+
+There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  The major use case for this is in a serverless functions environment,  or in a container based deployment, where the container being deployed is ephemeral.
```
lance, line 92:
```suggestion
There may be times where you will need to pre-populate the stats of the Circuit Breaker Status Object.  Primary use cases for this are in a serverless environment such as Knative or AWS Lambda, or any container based platform, where the container being deployed is ephemeral.
```
A couple of wording tweaks here, to emphasize that this isn't just for functions.

lance (commented):
Just a few wording suggestions

lib/circuit.js
```diff
@@ -138,7 +154,16 @@ class CircuitBreaker extends EventEmitter {
     this[VOLUME_THRESHOLD] = Number.isInteger(options.volumeThreshold)
       ? options.volumeThreshold : 0;
     this[WARMING_UP] = options.allowWarmUp === true;
-    this[STATUS] = new Status(this.options);
+
+    // The user can pass in a Status object ot inialize the Status/stats
+    if (this.options.status && this.options.status instanceof Status) {
+      // Do a check that this is a Status Object,
+      // should it error if it is not a Status Object
```
lance, line 161: I'm leaning towards yes. It seems wrong to just accept bad input and not do anything about it.

README.md
lholmquist: i knew that :)

lib/circuit.js
lholmquist: same,  i'll update to throw an error if the status option is not an instance of the Status class

lholmquist: updated

test/status-test.js
```diff
@@ -0,0 +1,159 @@
...
+    latencyTimes: []
+  };
+
+  const status = CircuitBreaker.newStatus({stats: prevStats});
+  t.equal(status instanceof Status, true, 'returns a new Status instance');
+
+  const stats = status.stats;
+  t.equal(stats.failures, 1, 'status reports 1 failure');
+  t.equal(stats.rejects, 1, 'status reports 1 reject');
+  t.equal(stats.fires, 1, 'status reports 5 fires');
```
helio-frota, line 31: status reports 5 fires, should be 1 fire?

helio-frota (changes requested):

test/status-test.js
lholmquist: yea,  i can't copy/paste very good

lholmquist: Updated

helio-frota (approved):

lance (approved):
Thanks for tackling this @lholmquist - looks good. :)

PR #574, opened 2021-05-07
feat: initialize the state of a breaker on creation

lholmquist:
This PR is a continuation of #568.  That last PR only allowed a user to prime the circuit breaker with the stats of the circuit.  This part adds the ability to prime the breaker with a starting state.  To make it easier for users to see there state, and ultimately save it some where, the `exportState` function has been added.

To create a Circuit with states, there is now a `state` parameter in the options.  Creating a new circuit might look like this:

```

const state = {
....
}


const breaker = new CircuitBreaker({state: state});



// to export the state

const breakerState = breaker.exportState();
```
I updated the README with some samples of how this would work

The only part i'm not thrilled with is but is probably ok, is that during initilization, if you prime a circuit to be open, then the code emits an "open" event so that the proper timers runs, but it also increments the status count of open, even though the protected function never ran.  It probably isn't even an issue since it only will increment the status by 1.

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -157,7 +157,12 @@ class CircuitBreaker extends EventEmitter {
 
     // The user can pass in a Status object to inialize the Status/stats
     if (this.options.status) {
-      this[STATUS] = new Status({ stats: this.options.status });
+      // Do a check that this is a Status Object,
+      if (!(this.options.status instanceof Status)) {
+        this[STATUS] = new Status(this.options.status);
+      } else {
+        this[STATUS] = this.options.status;
+      }
     } else {
       this[STATUS] = new Status(this.options);
     }
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -168,17 +168,22 @@ class CircuitBreaker extends EventEmitter {
       this[STATUS] = new Status(this.options);
     }
 
+    this[STATE] = CLOSED;
+
     if (options.state) {
       this[ENABLED] = options.state.enabled !== false;
       this[WARMING_UP] = options.state.warmUp || this[WARMING_UP];
-      this[CLOSED] = options.state.closed || false;
-      this[OPEN] = options.state.open || false;
+      // Closed if nothing is passed in
+      this[CLOSED] = options.state.closed !== false;
       // These should be in sync
       this[HALF_OPEN] = this[PENDING_CLOSE] = options.state.halfOpen || false;
+      // Open should be the opposite of closed,
+      // but also the opposite of half_open
+      this[OPEN] = !this[CLOSED] && !this[HALF_OPEN];
+
       this[SHUTDOWN] = options.state.shutdown || false;
 
     } else {
-      this[STATE] = CLOSED;
       this[PENDING_CLOSE] = false;
       this[ENABLED] = options.enabled !== false;
     }
@@ -254,10 +259,11 @@ class CircuitBreaker extends EventEmitter {
       CACHE.set(this, undefined);
     }
 
-     // Prepopulate the State of the Breaker
-    if (this[CLOSED]) {
-      // If the state is closed, we don't do anything?
-      this[STATE] = CLOSED;
+    // Prepopulate the State of the Breaker
+    if (this[SHUTDOWN]) {
+      this[STATE] = SHUTDOWN;
+      this.shutdown();
+    } else if (this[CLOSED]) {
       this.close();
     } else if (this[OPEN]) {
       // If the state being passed in is OPEN
@@ -266,11 +272,6 @@ class CircuitBreaker extends EventEmitter {
     } else if (this[HALF_OPEN]) {
       // Not sure if anythign needs to be done here
       this[STATE] = HALF_OPEN;
-    } else if (this[SHUTDOWN]) {
-      // IF the state of the circuit being passed in is SHUTDOWN
-      // Then i don't think we need to do anything either?
-      this[STATE] = SHUTDOWN;
-      this.shutdown();
     }
   }
 
@@ -412,16 +413,18 @@ class CircuitBreaker extends EventEmitter {
     return this[STATUS].stats;
   }
 
-  // This is only for convience
-  exportState () {
+  toJSON () {
     return {
-      enabled: this.enabled,
-      closed: this.closed,
-      open: this.opened,
-      halfOpen: this.halfOpen,
-      warmUp: this.warmUp,
-      pendingClose: this.pendingClose,
-      shutdown: this.isShutdown
+      state: {
+        name: this.name,
+        enabled: this.enabled,
+        closed: this.closed,
+        open: this.opened,
+        halfOpen: this.halfOpen,
+        warmUp: this.warmUp,
+        shutdown: this.isShutdown
+      },
+      status: this.status.stats
     };
   }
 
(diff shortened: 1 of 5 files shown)


lib/circuit.js
```diff
@@ -168,12 +168,24 @@ class CircuitBreaker extends EventEmitter {
       this[STATUS] = new Status(this.options);
     }
 
-    this[STATE] = CLOSED;
```
lance, line 171: I think this should stay as the default, and in the `if (options.state)` block, it can be updated if needed.

lib/circuit.js
```diff
@@ -241,6 +253,25 @@ class CircuitBreaker extends EventEmitter {
...
+      this.open();
+    } else if (this[HALF_OPEN]) {
+      // Not sure if anythign needs to be done here
+      this[STATE] = HALF_OPEN;
+    } else if (this[SHUTDOWN]) {
+      // IF the state of the circuit being passed in is SHUTDOWN
+      // Then i don't think we need to do anything either?
+      this[STATE] = SHUTDOWN;
+      this.shutdown();
+    }
```
lance, line 274: This should all be in the above `if (options.state) {}` block, shouldn't it? And could probably be simplified to something like this.

```suggestion
     if (options.state.shutdown) {
       this.shutdown();
     } else if (options.state.open) {
       this.open(); 
     }

     if (options.state.halfOpen) {
       this[STATE] = HALF_OPEN;
     }
```

lib/circuit.js
```diff
@@ -381,6 +412,19 @@ class CircuitBreaker extends EventEmitter {
...
+    return {
+      enabled: this.enabled,
+      closed: this.closed,
+      open: this.opened,
+      halfOpen: this.halfOpen,
+      warmUp: this.warmUp,
+      pendingClose: this.pendingClose,
+      shutdown: this.isShutdown
+    };
+  }
```
lance, line 426: Barring my comments above about field/behavior duplication, I like this. It got me thinking that a [`toJSON()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior) method might be useful. E.g.

```suggestion
  exportState () {
    return this.toJSON();
  }
  
  // ensures that when a circuit is serialized as JSON, it looks like this
  toJSON() {
    return {
      name: this.name,
      group: this.group,
      enabled: this.enabled,
      opened: this.opened,
      halfOpen: this.halfOpen,
      warmUp: this.warmUp,
      shutdown: this.isShutdown,
      status: this.status // obviously this would affect the ctor params
    };
  }
```
This also ensures that a circuit's state and status on import/export are in sync.

lance (commented):
Overall, I like it. I do have some thoughts on field duplication and other concerns that were not driving motivators with this change, but instead historical cruft that was just highlighted as the cruft that it is with this PR. I also wonder if it makes sense to completely separate state and status, so have made a suggestion about combining these two into a single `state` option that contains a status.

lib/circuit.js
```diff
@@ -168,12 +168,24 @@ class CircuitBreaker extends EventEmitter {
...
     }
 
-    this[STATE] = CLOSED;
+    if (options.state) {
+      this[ENABLED] = options.state.enabled !== false;
+      this[WARMING_UP] = options.state.warmUp || this[WARMING_UP];
+      this[CLOSED] = options.state.closed || false;
+      this[OPEN] = options.state.open || false;
+      // These should be in sync
+      this[HALF_OPEN] = this[PENDING_CLOSE] = options.state.halfOpen || false;
```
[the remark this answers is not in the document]
lholmquist: It seems that they are almost the same thing.  I did that quick test and it looks like we can probably remove pendingClose and just go with halfOpen

lib/circuit.js
```diff
@@ -168,12 +168,24 @@ class CircuitBreaker extends EventEmitter {
       this[STATUS] = new Status(this.options);
     }
 
-    this[STATE] = CLOSED;
+    if (options.state) {
+      this[ENABLED] = options.state.enabled !== false;
+      this[WARMING_UP] = options.state.warmUp || this[WARMING_UP];
+      this[CLOSED] = options.state.closed || false;
+      this[OPEN] = options.state.open || false;
```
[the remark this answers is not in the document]
lholmquist: > Why do we have both this[OPEN] and this[CLOSED]? 
Well, you wrote it  :)    i would have to look a little deeper to see if both are needed.  maybe that would be in a different PR though

We probably don't need to pass both in.  Since if it isn't closed, then it is Open.   I would opt for only using the closed parameter

lib/circuit.js
lholmquist: sounds good

lib/circuit.js
lholmquist: I'm pretty sure there was a reason for doing this way.  I think it was becuase there are event listeners and functions that need to get created first,  before this code gets executed.  but i'll have to go back and check that

lholmquist:
@lance i've made some updates. I added the `toJSON` method and removed the `exportState` function since they do the same thing.  I also added the `status.stats` to the `toJSON` function so it now exports something like this:
```
{
  state: { ... },
 status: { ... }
}
```

Which doesn't require any changes to the constructor.

I decided not to mess around with the pendingClose/HalfOpen stuff since that can be done in a future update since it is not user facing.

lance:
> I decided not to mess around with the pendingClose/HalfOpen stuff since that can be done in a future update since it is not user facing.

Probably a good idea.

lib/circuit.js
```diff
@@ -157,13 +157,7 @@ class CircuitBreaker extends EventEmitter {
...
     // The user can pass in a Status object to inialize the Status/stats
     if (this.options.status) {
-      // Do a check that this is a Status Object,
-      if (this.options.status instanceof Status) {
-        this[STATUS] = this.options.status;
-      } else {
-        // should it error if it is not a Status Object
-        throw new TypeError('Not a Status Object.  The status option must be an instance of Status');
-      }
+      this[STATUS] = new Status(this.options.status);
```
lance, line 160: Shouldn't this be like this?
```suggestion
      this[STATUS] = new Status({ stats: this.options.status });
```
https://github.com/nodeshift/opossum/blob/16354f2a7b1f7f30f62acc2481f6b0f2e9879cef/lib/status.js#L90-L92

lholmquist: derp,  yup

lib/circuit.js
```diff
@@ -158,22 +158,38 @@ class CircuitBreaker extends EventEmitter {
...
     if (this.options.status) {
       // Do a check that this is a Status Object,
-      if (this.options.status instanceof Status) {
-        this[STATUS] = this.options.status;
+      if (!(this.options.status instanceof Status)) {
+        this[STATUS] = new Status(this.options.status);
       } else {
-        // should it error if it is not a Status Object
-        throw new TypeError('Not a Status Object.  The status option must be an instance of Status');
+        this[STATUS] = this.options.status;
```
lance, line 164:
```js
if (this.options.status instanceof Status) {
  this[STATUS] = this.options.status;
} else if (this.options.status) {
  this[STATUS] = new Status( { stats: this.options.status } );
}
```

lance (approved):

PR #625, opened 2022-01-07
fix: breaker should emit a shutdown event when it is shutdown

### LINKED ISSUE #620: [BUG] 'shutdown' event is not able to be fired
## Node.js Version:
v16.11.1

## Operating System:
macOS Monterey (12.0.1)

## Steps to Produce Error:
It always fails.

```ts
/**
 * Shuts down this circuit breaker. All subsequent calls to the
 * circuit will fail, returning a rejected promise.
 * @returns {void}
 */
shutdown () {
  this.disable();
  this.removeAllListeners();
    
  // intentionally omitted

  /**
   * Emitted when the circuit breaker has been shut down.
   * @event CircuitBreaker#shutdown
   */
   this.emit('shutdown');
}
```

With current implementation of `shutdown`, `shutdown` event can never be fired since `shutdown` removes all listeners from the beginning.

`this.emit('shutdown')` must be called before `this.removeAllListeners()`.

lholmquist:
fixes #620

currently, when shutdown is called, it removes all the listeners before it emits the shutdown event, this moves the call to emit the shutdown event before the listeners are removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ Here are the events you can listen for.
   * `fallback` - emitted when the breaker has a fallback function and executes it
   * `semaphoreLocked` - emitted when the breaker is at capacity and cannot execute the request
   * `healthCheckFailed` - emitted when a user-supplied health check function returns a rejected promise
-  * `shutdown` - emitted when the breaker shutsdown
+  * `shutdown` - emitted when the breaker shuts down
 
 Handling events gives a greater level of control over your application behavior.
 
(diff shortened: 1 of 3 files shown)


lholmquist:
@leejh3224 i think this should do it

README.md
```diff
@@ -259,6 +259,7 @@ Here are the events you can listen for.
   * `fallback` - emitted when the breaker has a fallback function and executes it
   * `semaphoreLocked` - emitted when the breaker is at capacity and cannot execute the request
   * `healthCheckFailed` - emitted when a user-supplied health check function returns a rejected promise
+  * `shutdown` - emitted when the breaker shutsdown
```
lance, line 262:
```suggestion
  * `shutdown` - emitted when the breaker shuts down
```
Nit

lance (approved):
Teeny tiny nit

Issue #673, opened 2022-08-24
Add AbortController support


raytung:
Hey team,

This is a feature request to add support for `AbortController` 
- https://developer.mozilla.org/en-US/docs/Web/API/AbortController
- https://nodejs.org/docs/latest-v15.x/api/globals.html#globals_class_abortcontroller

With `AbortController`, Opossum should be able to send signal to the underlying method to abort the on going request. This is useful in conjunction with Opossum's timeout feature, where Opossum will fail the on going request but also allow the on going HTTP request to be properly aborted instead of hanging around in the background 

- https://nodejs.org/api/http.html#httprequestoptions-callback
- https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#aborting_a_fetch

lance:
Fixed in https://github.com/nodeshift/opossum/pull/674

PR #674, opened 2022-08-24
feat: adds `AbortController` support

raytung:
Addresses https://github.com/nodeshift/opossum/issues/673

This PR adds `AbortController` support in Opossum. This is useful in
conjunction with the timeout feature, where Opossum will signal upon
timeout and allow users to cleanly abort their ongoing requests.

See links:
- https://developer.mozilla.org/en-US/docs/Web/API/AbortController
- https://nodejs.org/api/globals.html#class-abortcontroller
- https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#aborting_a_fetch
- https://nodejs.org/api/http.html#httprequesturl-options-callback

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,44 @@ breaker.fire(x, y)
   .catch(console.error);
 ```
 
+### AbortController support
+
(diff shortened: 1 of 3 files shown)


lance (approved):
@raytung this is a great addition - thanks!
/lgtm

PR #680, opened 2022-10-03
chore: update CI to include build+test:headless

lance:
Because some of the dependabot / snyk PRs could update dependencies like webpack and friends, these checks should be included for pull requests.


lance:
Huh looks like this caught itself. I was using Node.js 16 for this PR, but apparently a change in 18 breaks the browser build.

PR #682, opened 2022-10-11
chore: clean up some dependenices and lint

lance:
This commit removes stray dependencies on codacy and coveralls, and removes the dependency on tap-spec because of security vulnerabilities. It has been replaced with `faucet` which seems to be more regularly updated.

I've also corrected a couple of linting errors and fixed the `npm lint` command.


lance:
Closing and reopening to see if newly installed codecov app will generate a coverage report

Issue #758, opened 2023-05-19
Is there a reason you are mixing rest parameters and Array.prototype.slice calls in the same functions?


jdmarshall:
I'm trying to switch an app to opossum from hystrix and I'm seeing some perf regressions, which I found surprising considering how complex the hystrix internals are.

Looking for easy changes that could improve performance and ran into situations such as:

https://github.com/nodeshift/opossum/blob/main/lib/circuit.js#L558

On first read this looks like a guarded copy of inputs in order to prevent other callers from modifying the data, however the args object is first passed to an event emitter and then used afterward. Which means it doesn't achieve that goal and still involves call(). In fact there are a number of functions that use .apply() or .call() and also ...rest. So I'm curious if this has a purpose I'm not familiar with, or is just an incomplete refactor of old code to use ES6 features. 

All of these have clarity issues, but some also have performance issues, and I'm wondering where or if to spend my time.

lholmquist:
@jdmarshall thanks for the find.  i'd say that the reason for both was because of an incomplete refactor.

PR #765, opened 2023-05-27
Reduce apply() calls to improve call times.

jdmarshall:
This brings opossum.fire() roughly in line with hystrixjs.execute()

Reduces fire() call time by about 50%, and fallback call time by 30%

    fire() x 134,900 ops/sec ±42.25% (21 runs sampled)
    fire() error x 68,030 ops/sec ±31.60% (8 runs sampled)
    
    fire() x 286,626 ops/sec ±28.73% (56 runs sampled)
    fire() error x 99,637 ops/sec ±2.48% (10 runs sampled)

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -519,9 +519,7 @@ class CircuitBreaker extends EventEmitter {
   fallback (func) {
     let fb = func;
     if (func instanceof CircuitBreaker) {
-      fb = function () {
-        return func.fire.apply(func, arguments);
-      };
+      fb = (...args) => func.fire(...args);
     }
     this[FALLBACK_FUNCTION] = fb;
     return this;
@@ -550,7 +548,7 @@ class CircuitBreaker extends EventEmitter {
    * @fires CircuitBreaker#semaphoreLocked
    */
   fire (...args) {
-    return this.call.apply(this, [this.action].concat(args));
+    return this.call(this.action, ...args);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lib/circuit.js
```diff
@@ -582,7 +580,8 @@ class CircuitBreaker extends EventEmitter {
       const err = buildError('The circuit has been shutdown.', 'ESHUTDOWN');
       return Promise.reject(err);
     }
-    const args = Array.prototype.slice.call(rest);
+
+    const args = rest.slice();
```
jdmarshall, line 584: rest arguments become [] when no arguments provided, so Array.prototype.slice is guarding nothing.

jdmarshall:
This addresses questions asked and answered in #758

lholmquist (approved):

lance (approved):
/lgtm

lib/circuit.js
jdmarshall: Sounds about right

jdmarshall:
Talking it out with some coworkers this morning, I mentioned I was a little surprised by how effective this PR was. I was expecting about 2/3rds of this improvement at the outside.

What I'm thinking is that the apply/rest performance discrepancy may have increased in Node 16, when V8 changed how it handled function arguments (specifically to speed up calls with the 'wrong' number of args). I looked at this when rest parameters were first introduced and while it was helpful, it wasn't 'drop everything' level of performance gap.

At a guess, I'm thinking the average overhead for regular calls went down compared to apply(), anonymous functions versus arrow functions, functions that use arguments, or some combination of the three.

lholmquist:
just released 8.0.1

PR #769, opened 2023-05-30
chore: update github actions for coveralls

lance:
Ensures that the coveralls build is not closed until both `lts` and `current` jobs complete.


lance:
@lholmquist I think this is the root of the problem with coveralls in https://github.com/nodeshift/opossum/pull/765

richardlau (approved):

lholmquist (approved):

Issue #798, opened 2023-07-29
Feature request: add an option to disable stats snapshots


gjethwani:
Hello! I had a small feature request

**Motivation**

As I understand it, every `bucket_interval` (1s by default), opossum emits a snapshot of the stats in the `Status` class

```
this[SNAPSHOT_INTERVAL] = setInterval(_ => this.emit('snapshot', this.stats), bucketInterval);
```

My app doesn't actually listen to these stats and the extra call to `setInterval` clogs up the memory

**Solution**

Could we have a configurable option to disable the stats snapshots, thereby removing the extra call to `setInterval`?

gjethwani:
Created a PR here: https://github.com/nodeshift/opossum/pull/799

gjethwani:
Corresponding PR in DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66220

lholmquist:
just published 8.1.0

PR #799, opened 2023-07-29
feat: add option to disable stats snapshots

gjethwani:
Hello! I had a small feature request

Motivation

As I understand it, every bucket_interval (1s by default), opossum emits a snapshot of the stats in the Status class

this[SNAPSHOT_INTERVAL] = setInterval(_ => this.emit('snapshot', this.stats), bucketInterval);
My app doesn't actually listen to these stats and the extra call to setInterval clogs up the memory

Solution

Could we have a configurable option to disable the stats snapshots, thereby removing the extra call to setInterval?

Issue here: https://github.com/nodeshift/opossum/issues/798

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -158,6 +158,7 @@ class CircuitBreaker extends EventEmitter {
     this.options.cacheTTL = options.cacheTTL ?? 0;
     this.options.cacheGetKey = options.cacheGetKey ??
       ((...args) => JSON.stringify(args));
+    this.options.enableSnapshots = options.enableSnapshots !== false
 
     // Set default cache transport if not provided
     if (this.options.cache) {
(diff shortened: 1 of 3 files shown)


gjethwani:
Corresponding PR in DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66220

lholmquist (changes requested):
This looks pretty good.  can you please add a test for this

gjethwani:
Thanks @lholmquist , just added some tests. Could you please give it a re-review?

gjethwani:
Also, would you like me to take care of the PR in DefinitelyTyped or is that something you would do when you release a new version?

lance (approved):
Looks good - thank you for the contribution!

PR #825, opened 2023-10-04
fix: fix bug where options couldn't be set to zero.

### LINKED ISSUE #824: Setting some options to `0` doesn't work (it still uses defaults)
## Problem
I noticed when writing tests that setting some `options` to `0` is not working as I expected.

An example is `errorThresholdPercentage`, which I wanted to set to `0` in a test to make it fail immediately. But setting it to `0` will result in the default value `50` being used!

Is that the expected behaviour?

---
The reason why this happens is in how the defaults are set in the `CircuitBreaker` constructor.
```
this.options.errorThresholdPercentage =
      options.errorThresholdPercentage || 50;
```
https://github.com/nodeshift/opossum/blob/main/lib/circuit.js#L157C5-L158C46

Since `0` is considered `false` it will take the `50`.

Similar behaviour is also present for the options `resetTimeout`, `rollingCountTimeout`, `rollingCountBuckets`.

lholmquist:
when setting things like resetTimeout and errorThresholdPercentage to 0, the default value was being used since zero is a false-y value.  Adding the [nullish coalesce operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#assigning_a_default_value_to_a_variable) fixes it 

fixes #824

### DIFF
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -151,13 +151,12 @@ class CircuitBreaker extends EventEmitter {
   constructor (action, options = {}) {
     super();
     this.options = options;
-    this.options.timeout =
-      options.timeout === false ? false : options.timeout || 10000;
-    this.options.resetTimeout = options.resetTimeout || 30000;
+    this.options.timeout = options.timeout ?? 10000;
+    this.options.resetTimeout = options.resetTimeout ?? 30000;
     this.options.errorThresholdPercentage =
-      options.errorThresholdPercentage || 50;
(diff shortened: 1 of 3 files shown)


lholmquist:
@ollelindeman this look good?

ollelindeman:
> @ollelindeman this look good?

Look good to me! Thanks for the fast reply and action! 👍 🚀

PR #892, opened 2024-10-23
feat: renewing the AbortController when the circuit enters the 'halfClose' or 'close' state

### LINKED ISSUE #861: AbortController break fetch after first timeout because aborted signal can't be re-used
## Node.js Version:

```
↪ node --version
v20.12.2
```

## Operating System:

```
↪ uname -a
Darwin Mac-WP9QH543YX 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64
```

## The problem:

`AbortSignal` both [stores it's state](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/aborted) and [provides an `abort` event](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/abort_event) to signal when abort is happening.

And `AbortController` newer updates it's `signal`, leaving it in `aborted` state forever:

```
> abortController = new AbortController()
AbortController { signal: AbortSignal { aborted: false } }
> abortController.abort()
undefined
> abortController
AbortController { signal: AbortSignal { aborted: true } }
```

It looks like `undici` fetch uses both `aborted` property and `abort` event to skip sending the request if it is already aborted.

So, once the action times out for the first time, it switches `AbortController` into `aborted` state and all subsequent fetch calls just abort immediately.

## Minimal reproducible example:

```js
const { describe, it, before, after, afterEach, mock } = require('node:test');
const assert = require('node:assert/strict');
const http = require('node:http');

const CircuitBreaker = require('opossum');

describe('Circuit breaker with abort controller', () => {
    const abortController = new AbortController();
    const get = mock.fn((timeout) => {
        const { port } = server.address();
        const { signal } = abortController;
        const headers = { 'x-timeout': timeout };
        return fetch(`http://localhost:${port}`, { signal, headers });
    });
    const breaker = new CircuitBreaker(get, {
        abortController,
        timeout: 1000,
        volumeThreshold: 3
    });

    const server = http.createServer((req, res) => {
        const timeout = Number(req.headers['x-timeout']);
        setTimeout(() => res.end(`delayed for ${timeout}ms`), timeout);
    });

    before(() => server.listen());
    after(() => server.close());
    afterEach(() => get.mock.resetCalls());

    it('Should work', async () => {
        const response = await breaker.fire(100);
        const text = await response.text();
        assert.equal(text, 'delayed for 100ms');
    });

    it('Should timeout and abort request', async () => {
        await assert.rejects(() => breaker.fire(2000), {
            message: 'Timed out after 1000ms'
        });
        await assert.rejects(() => get.mock.calls[0].result, {
            message: 'This operation was aborted'
        });
    });

    it('Should work again', async () => {
        try {
            await breaker.fire(100);
        } catch ({ name, message }) {
            throw Object.assign(new Error(message), { name });
        }
    });
});
```

```
↪ node --test
▶ Circuit breaker with abort controller
  ✔ Should work (153.779083ms)
  ✔ Should timeout (1008.148333ms)
  ✖ Should work again (1.391459ms)
    Error [AbortError]: This operation was aborted
        at TestContext.<anonymous> (/Users/leonbesc/Documents/viaplay/opossum-abort-controller-issue/test.js:42:33)
        at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
        at async Test.run (node:internal/test_runner/test:640:9)
        at async Suite.processPendingSubtests (node:internal/test_runner/test:382:7)
```

## Possible ways to fix this:

It seems we cannot re-use the same `AbortController` and `AbortSignal` pair for all requests since `AbortSignal` preserves its `aborted` state.

I can only see one way to fix this problem - use new `AbortSignal` for every requests (see #780).

For example, `opossum` could pass the `signal` object as the first argument to the action handler.

WillianAgostini:
Introducing a new option to the CircuitBreaker: `autoRenewAbortController`. When enabled, this flag automatically recreates the AbortController whenever the CircuitBreaker state transitions to 'halfOpen' or 'closed'. This ensures the controller is reset and ready for reuse in subsequent state changes.

Closes #861

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,44 @@ breaker.fire(abortController.signal)
   .catch(console.error);
 ```
 
+### Auto Renew AbortController
+
+The `autoRenewAbortController` option allows the automatic renewal of the `AbortController` when the circuit breaker transitions into the `halfOpen` or `closed` states. This feature ensures that the `AbortController` can be reused properly for ongoing requests without manual intervention.
+
+```javascript
+const CircuitBreaker = require('opossum');
+const http = require('http');
+
+function asyncFunctionThatCouldFail(abortSignal, x, y) {
+  return new Promise((resolve, reject) => {
+    http.get(
+      'http://httpbin.org/delay/10',
+      { signal: abortSignal },
+      (res) => {
+        if(res.statusCode < 300) {
+          resolve(res.statusCode);
+          return;
+        }
+
+        reject(res.statusCode);
+      }
+    );
+  });
+}
+
+const abortController = new AbortController();
+const options = {
+  autoRenewAbortController: true,
+  timeout: 3000, // If our function takes longer than 3 seconds, trigger a failure
+};
+const breaker = new CircuitBreaker(asyncFunctionThatCouldFail, options);
+
+const signal = breaker.getSignal();
+breaker.fire(signal)
+  .then(console.log)
+  .catch(console.error);
+```
+
 ### Fallback
 
 You can also provide a fallback function that will be executed in the
--- a/lib/circuit.js
+++ b/lib/circuit.js
@@ -107,6 +107,11 @@ Please use options.errorThresholdPercentage`;
  * This option allows you to provide an EventEmitter that rotates the buckets
  * so you can have one global timer in your app. Make sure that you are
  * emitting a 'rotate' event from this EventEmitter
+ * @param {boolean} options.autoRenewAbortController Automatically recreates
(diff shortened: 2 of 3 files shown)


lholmquist (commented):
Could you also update the README in the abort controller section mentioning this new feature?

WillianAgostini:
> Could you also update the README in the abort controller section mentioning this new feature?

I've added a new commit that updates the README in the abort controller section

Issue #894, opened 2024-10-25
Behaviour change in fire statistics in v8.2.0


ullumullu:
I've discovered a small regression (?) which seems to have been introduced with https://github.com/nodeshift/opossum/pull/877 affecting disabled breakers.

**Before**

```
this.emit('fire', args);
```

The [fire event](https://github.com/nodeshift/opossum/pull/877/files#diff-9d0c1cf88ea5d623af8143c3b2c2fc7e62f92592388891ee8f6071f24abc4264R638 ) was triggered for disabled breakers as well.

**After**

Moving the [disablement check](https://github.com/nodeshift/opossum/pull/877/files#diff-9d0c1cf88ea5d623af8143c3b2c2fc7e62f92592388891ee8f6071f24abc4264R623) before the fire event changes the existing behaviour for disabled breakers.

---

I am not sure if this change was intentional - however it could lead to unwanted side effects for disabled breakers. Moving the emit call before the disablement check would revert the behaviour to the existing status.

![Image](https://github.com/user-attachments/assets/b1c653b4-9840-469a-a8e8-b357ac0fedf2)

PR #895, opened 2024-10-25
fix: disabled breaker emits fire events

ullumullu:
To revert to previous behaviour up to v8.1.4.

https://github.com/nodeshift/opossum/issues/894


lholmquist:
can you add a test that makes sure to test this behaviour?

lholmquist:
I noticed this was a Draft,  it looks good, was there something else you wanted to add, or just hit the wrong button :)

lholmquist (approved):

PR #896, opened 2024-10-25
Bugfix for 8.20.0

daan944:
Bugfix: coalesce re-used empty cachekey when cache was not used.
Bugfix: emit fire when circuitbreaker is disabled.


daan944:
Of course!

Issue #904, opened 2024-11-04
Incorrect default value of maxEntries for MemoryCache


IvanPenga:
Added in opossum version 8.2.0, if cacheSize property is not provided to MemoryCache constructor, it will default to 2 ^ 24 - 1, which is 21, because caret (^) operator is bitwise operator, not exponentiation operator. This should be replaced with ** operator, or Math.pow function.

daan944:
That's my bad, sorry. Fixed it in [PR](https://github.com/nodeshift/opossum/pull/906)

lholmquist:
fix is in the newly released 8.3.1

PR #906, opened 2024-11-08
Fixed: Incorrect default value of maxEntries for MemoryCache #904

daan944:
See https://github.com/nodeshift/opossum/issues/904


lholmquist:
could you add a test that checks that this value is correct?

PR #908, opened 2024-11-20
feature: Coalesce reset options, reset coalesce based on result. (error, success, timeout)

daan944:
After implementing coalescing in our software, we noticed it would be handy to better finetune coalescion behavior. 

So added the option to reset coalescion cache at different stages: error, success, timeout. This can be used as sometimes you'd like to retry after an error/timeout, or would like to limit cache TTL without limiting coalescion TTL (e.g. a slow action to fetch data you do want to execute as often as possible but never twice at the same time).

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,7 +459,45 @@ The code that is summing the stats samples is here:
 
 ### Coalesce calls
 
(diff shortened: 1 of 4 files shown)


lholmquist:
could you also update the readme with this new functionally

daan944:
Done!

lholmquist (approved):

lholmquist:
released as 8.4.0 on npm

PR #958, opened 2026-08-04
fix: reject non-finite numeric options instead of silently disabling features

### LINKED ISSUE #934: Null coalescing operator should not be used for setting default options
The circuit breaker constructor [is using](https://github.com/nodeshift/opossum/blob/3015ba3c5134a7010d16bb50b29e7a7a7d91aa7f/lib/circuit.js#L179-L180) null coalesce operators for some settings like `timeout`. However this can cause (and caused) unexpected behaviour down the line. 

For instance `timeout` setting is falsey checked before use [here](https://github.com/nodeshift/opossum/blob/3015ba3c5134a7010d16bb50b29e7a7a7d91aa7f/lib/circuit.js#L732). However due to null coalesce, `NaN` is a valid `timeout` setting and can be set if a consumer of the library is parsing an integer.

I guess you could argue `NaN` setting is working as expected (not timing out) but `setTimeout` treats `NaN` as `0`.

1991santhu:
Fixes #934.

## The problem

The constructor uses `??` for numeric options, which only substitutes the default for `null`/`undefined`. A value that is number-typed but unusable — the `NaN` you get from `parseInt(process.env.SOMETHING_UNSET)` — passes through as though the caller supplied it.

Nothing downstream catches it, and the result is worse than the issue describes. The timeout call site guards on truthiness:

```js
if (this.options.timeout) {
  timeout = setTimeout(..., this.options.timeout);
}
```

`NaN` is falsy, so `setTimeout` is never scheduled — the breaker ends up with **no timeout at all**, rather than an immediate one.

There is a second path: `coalesceTTL` inherits the `NaN` via `?? this.options.timeout`, and its use site compares `this.options.coalesceTTL > 0`, which is false for `NaN`, so coalescing is silently disabled too.

In both cases a misconfigured breaker looks healthy and quietly stops bounding anything — which is the specific outcome the library exists to prevent.

## The change

Validate at construction, so a misparsed value fails loudly at startup instead of degrading silently at runtime. `Number.isFinite` covers `NaN` and both infinities in one check.

This follows existing precedent in the same constructor — `capacity` already validates with `Number.isInteger` rather than trusting `??`.

Applied to `timeout`, `resetTimeout`, `errorThresholdPercentage`, `rollingCountTimeout`, `rollingCountBuckets`, `cacheTTL` and `coalesceTTL`.

## Compatibility

- `timeout: false` is passed through unchanged — it is the documented way to disable the timeout, and the existing suite covers it. (I initially missed this; the test suite caught it.)
- `0` remains valid everywhere, which is why `||` is not an acceptable substitute for `??` here.
- Omitted options still receive their defaults.

The one behaviour change is that a value which is currently accepted and silently ignored now throws. That is deliberate — anyone passing `NaN` today has a breaker that is not doing its job — but I am happy to switch to falling back to the default with a warning if maintainers prefer a non-throwing path.

## Testing

New `test/numeric-options-test.js` covers rejection of `NaN`, `Infinity` and non-numeric types across all seven options, and asserts that `0`, `false`, valid numbers and omitted options all still behave as before.

Full suite passes locally: **569/569**, lint clean.


lholmquist:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/nodeshift-agg/pr-checks/2ca0c5b3-b7ac-44ee-b3c2-c44ef6ff7139?refs=55d50a94b23e2bac1cc43d4b7e9d819df662f67f&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

